PR #5, opened 2020-07-21
Allow API to run on different interface/port

nikita-vanyasin:
Part of DEV-1589

- Some unused functions removed (commit №1) 
- Server god-object split into 3 entities: ClientListener (manages incoming client connections), APIListener (manages API requests) and SessionRepository (manages active client sessions). SessionRepository later can be extracted into interface allowing to implement persistent storage. Most of ClientListener code just moved from handler.go
- ErrorsCollector type added similar to what we have in cagent and frontman but with thread-safe consideration

That's a lot of changes, please let me know if any design decisions are unclear or maybe naming can be improved.


server/client_listener.go
```diff
@@ -0,0 +1,306 @@
...
+	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/jpillora/requestlog"
+	"golang.org/x/crypto/ssh"
+
+	chshare "github.com/cloudradar-monitoring/rport/share"
+)
+
+type ClientListener struct {
```
martinlindhe, line 22: Since both APIListener and ClientListener is listening for connections/commands from clients, maybe ClientListener should be called WSListener?

nikita-vanyasin: I see there maybe some confusion because of both types are listening to some incoming HTTP connections.
I think WSListener is no good because WebSockets is about implementation, not the purpose of this type. 

I think the purpose of the 'apilistener' is to handle API requests and 'clientlistener' is to handle connections from rport client applications.

How about `APIHandler` and `ClientHandler`?

martinlindhe: I read APIListener as "here is the rpc api implementation" when reading the patch, and the ClientListener serves over websocket, but I see your point.
Let's just leave the current names then, because I can't come up with a better name.

PR #7, opened 2020-07-29
Add authorization checks for API and API documentation

nikita-vanyasin:
DEV-1611

- API documentation added (swagger 2.0), see file api-doc.yml
- new CLI options --api-auth and --api-jwt-secret
- allow HTTP-basic authorization
- allow Bearer authorization using JWT tokens
- information about token expiration stored in-memory, meaning server restart will shutdown active api user sessions.  Later we will add support for persistent storage. For this purpose public methods of storage managers (APISessionStorage and ClientSessionStorage) were unified.
- initial token expiration still can be specified in /login requests with token-lifetime option. Max value is 90 days (in seconds). Default: 10 min.
- each request with `Authorization: Bearer XXX` header increases token lifetime (+10 min). Token does not hold expiration date meaning it remains unchanged.

### DIFF
--- a/server/bearer_auth.go
+++ b/server/bearer_auth.go
@@ -1,0 +1,74 @@
+package chserver
+
+import (
+	"math/rand"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/dgrijalva/jwt-go"
+)
+
+const (
+	maxTokenLifetime     = 90 * 24 * time.Hour
+	defaultTokenLifetime = 10 * time.Minute
+)
+
+func (al *APIListener) createAuthToken(lifetime time.Duration) (string, error) {
+	claims := jwt.StandardClaims{
+		Id: strconv.FormatUint(rand.Uint64(), 10),
+	}
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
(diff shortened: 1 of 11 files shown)


server/bearer_auth.go
```diff
@@ -0,0 +1,74 @@
+package chserver
+
+import (
+	"math/rand"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/dgrijalva/jwt-go/v4"
```
martinlindhe, line 10: I think it's better to use jwt-go v3 that is in maintenance mode, rather than v4 which is still unreleased.

From their readme:
> If you depend on this library and don't want to be interrupted, I recommend you use your dependency mangement tool to pin to version 3.

nikita-vanyasin: yeah, I've tried using v3 but there was some problem with go mod compatibility.
I've tried again with `go mod tidy` and now it works with v3 just fine.

nikita-vanyasin: Not sure about +incompatible flag though. I'll try some more things from here: https://github.com/dgrijalva/jwt-go/issues/294

nikita-vanyasin: Seems all good though. `+incompatible` marks that package wasn't using go versioning guidelines previously.

PR #9, opened 2020-07-29
Add option to serve static files when running API server

nikita-vanyasin:
Part of DEV-1589


nikita-vanyasin:
Rebased & fixed conflicts

PR #11, opened 2020-07-30
better identification of clients

martinlindhe:
* Give clients the option to use a persistent ID, a name, and tags.
* Readout some information about the host so the customer can identify the host on the server. The external IP address is useless if the customer connects many hosts behind the same NAT.
* Reject client connect attempts if another client is already using the id.
* Add 'os', 'hostname', 'ipv4', 'ipv6' to /api/v1/sessions

DEV-1591


martinlindhe:
@nikita-vanyasin 
I notice the build is failing on linux_ppc64. Is this a platform we should be building for? We are not building it for cagent or frontman.

nikita-vanyasin:
> @nikita-vanyasin
> I notice the build is failing on linux_ppc64. Is this a architecture we should be building for?
> It seems we are building many architectures for rport that is not enabled on cagent or frontman.
>

Goreleaser build configuration was copied from jpillora/chisel. I think if adding support for `linux_ppc64` platform requires additional code, we can drop it. It wasn't on a list of platforms, required by initial Proof of Concept description (DEV-1178).

cmd/rport/main.go
```diff
@@ -107,7 +107,14 @@ var clientHelp = `
     Can be used multiple times. (e.g --header "Foo: Bar" --header "Hello: World")
 
     --hostname, Optionally set the 'Host' header (defaults to the host
-    found in the server url).
+	found in the server url).
```
nikita-vanyasin, line 110: Not sure, why extra white-space added, new CLI options misaligned

server/client_session.go
```diff
@@ -88,3 +94,16 @@ func (c *ClientSession) removeTunnel(t *Tunnel) {
 	}
 	c.Tunnels = result
 }
+
+func (c *ClientSession) banner() string {
```
nikita-vanyasin, line 98: I think naming this function with Upper-case letter is better because it's called outside of client_session.go
I was thinking about further splitting ./server pkg into smaller sub-packages, so I've named all other methods according to this idea.

share/config.go
```diff
@@ -6,8 +6,15 @@ import (
 )
 
 type Config struct {
-	Version string
-	Remotes []*Remote
+	Version  string
+	ID       string
+	Name     string
+	OS       string //uname
```
nikita-vanyasin, line 12: that comment seems pretty useless, at least I could not understand why it's required here :)

share/uname.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if err := cmd.Wait(); err != nil {
+		return buf.Bytes(), err
+	}
+
+	return buf.Bytes(), nil
+}
+
+var invoke Invoker = Invoke{}
```
nikita-vanyasin, line 45: How about moving `Invoke` and `Invoker` type definitions to a separate file? It can come in handy for other commands, not only `uname`

martinlindhe:
> > @nikita-vanyasin
> > I notice the build is failing on linux_ppc64. Is this a architecture we should be building for?
> > It seems we are building many architectures for rport that is not enabled on cagent or frontman.
> 
> Goreleaser build configuration was copied from jpillora/chisel. I think if adding support for `linux_ppc64` platform requires additional code, we can drop it. It wasn't on a list of platforms, required by initial Proof of Concept description (DEV-1178).

Yea, it caused a build failure in gopsutil. I'll drop linux_ppc64 then.

cmd/rport/main.go
martinlindhe: Accidental tabs sneaked in. Good catch

server/client_session.go
martinlindhe: Sure, fixed

share/config.go
martinlindhe: Removed

share/uname.go
martinlindhe: Sure. This was just copied from frontman/borrowed.go

martinlindhe:
Current set of feedback should now be addressed.

PR #13, opened 2020-07-31
Documentation

thorstenkramm:
updated the quick start guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ For example
 ```
 CLIENTID=2ba9174e-640e-4694-ad35-34a2d6f3986b
 LOCAL_PORT=4000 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -295,6 +295,27 @@ Using `curl -u admin:foobaz -s http://localhost:3000/api/v1/sessions|jq` again c
     ]
 ```
 
+The rport client is not limited to establish tunnels only to the system it runs on. You can use it as a jump host to create tunnels to foreign systems to.
```
nikita-vanyasin, line 298:
```suggestion
The rport client is not limited to establish tunnels only to the system it runs on. You can use it as a jump host to create tunnels to foreign systems too.
```

PR #16, opened 2020-08-07
Drop support for address filtering in authfile JSON

nikita-vanyasin:
Part of DEV-1650


nikita-vanyasin:
canceling the PR, a new one will be created on a different branch

PR #17, opened 2020-08-10
Fix dependencies list with `go mod tidy`

nikita-vanyasin:
Looks like it wasn't updated properly after installing gopsutil previosly

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,9 +3,11 @@ module github.com/cloudradar-monitoring/rport
 go 1.13
 
 require (
+	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
 	github.com/andrew-d/go-termutil v0.0.0-20150726205930-009166a695a2 // indirect
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/fsnotify/fsnotify v1.4.9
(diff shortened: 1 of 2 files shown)


martinlindhe:
Maybe we can add a CI step for this if it's important?

nikita-vanyasin:
> Maybe we can add a CI step for this if it's important?

The actual problem comes from the fact, that `go get` or `go build` are adding dependencies only for current GOOS/GOARCH.

From go mod [docs](https://github.com/golang/go/wiki/Modules):
> go mod tidy also ensures your current go.mod reflects the dependency requirements for all possible combinations of OS, architecture, and build tags (as described here). In contrast, other commands like go build and go test only update go.mod to provide the packages imported by the requested packages under the current GOOS, GOARCH, and build tags (which is one reason go mod tidy might add requirements that were not added by go build or similar).

There are many discussions on this topic on Golang repo https://github.com/golang/go/issues/26571 with no real advice on how to sync all possible OS/ARCH combinations.

So, I'm still unsure if this is a good practice to add CI step. I think it's better to make all changes to dependencies list explicitly.

martinlindhe:
For a CI step I mean automated detection for when "go mod tidy" was not run.
* check out git master
* go mod tidy
* fail if git tree is dirty

PR #20, opened 2020-08-13
Show fingerprint and connect URL on API /status response object. New CLI arg allows to specify custom URL

nikita-vanyasin:
DEV-1643


nikita-vanyasin:
Rebased onto master, conflicts fixed

PR #22, opened 2020-08-17
Add handler for panicking HTTP handlers

nikita-vanyasin:
Now API will return InternalServerError instead of failing quietly

### DIFF
--- a/server/api_listener.go
+++ b/server/api_listener.go
@@ -4,6 +4,7 @@ import (
 	"crypto/subtle"
 	"fmt"
 	"net/http"
+	"runtime"
 
 	"github.com/gorilla/mux"
 	"github.com/jpillora/requestlog"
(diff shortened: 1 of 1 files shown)


server/api_listener.go
```diff
@@ -75,6 +76,16 @@ func (al *APIListener) Close() error {
 }
 
 func (al *APIListener) handleAPIRequest(w http.ResponseWriter, r *http.Request) {
+	defer func() {
```
martinlindhe, line 79: What part is causing the panic in the first place?

nikita-vanyasin: Currently there is no code that causes panic (at least known to me).
`recover()` handler added to allow enhanced logging and correct HTTP response in case there will be such unhandled error in future.
This code is pretty similar to what many Golang HTTP routers provide as "recovery middleware". e.g.: [gorilla](https://pkg.go.dev/github.com/gorilla/handlers?tab=doc#RecoveryHandler), [gin](https://github.com/gin-gonic/gin#blank-gin-without-middleware-by-default)

Not sure how stack trace can show config values. Only function names and memory addresses are printed. It also prints only to logfile/stdout, not to a HTTP response.

martinlindhe: Okay, but is there any reason to add code that's not in use? I don't think gorilla or gin is a comparable use case to ours since they have to deal with unknown users of their code which may or may not panic.
We shouldn't panic and thus shouldn't need to recover().

nikita-vanyasin: I think this is pretty the same use-case: we have a bunch of HTTP handlers for API endpoints. This endpoints can be triggered by any rportd user similarly to any HTTP API constructed with gin or gorilla.
Having a fail-over handler is pretty basic requirement for any web-server.
The point is to make sure end-user will get a nice HTTP error code (500) instead of empty response if rportd code will have a bug in a future (which is inevitable :smile:  ).

martinlindhe: If you feel this is necessary, then lets go ahead and merge it.

PR #23, opened 2020-08-17
Add option to specify tunnel ACL (range of allowed IPs)

nikita-vanyasin:
DEV-1636

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -228,13 +228,23 @@ func (al *APIListener) handlePutSessionTunnel(w http.ResponseWriter, req *http.R
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


server/api_listener.go
```diff
@@ -75,6 +76,16 @@ func (al *APIListener) Close() error {
 }
 
 func (al *APIListener) handleAPIRequest(w http.ResponseWriter, r *http.Request) {
```
martinlindhe, line 78: The panic handling was also submitted in https://github.com/cloudradar-monitoring/rport/pull/22/files

nikita-vanyasin: Rebase done

PR #27, opened 2020-08-27
Migrate CLI parsing to cobra

nikita-vanyasin:
That will allow to implement config files support in a clean way using spf13/viper.

DEV-1673


martinlindhe (approved):
Nice cleanup!

PR #28, opened 2020-08-28
DEV-1642: Store and read clients from file

m-terel:


cmd/rportd/main.go
```diff
@@ -110,6 +124,24 @@ func main() {
 	version := flag.Bool("version", false, "")
 	e := flag.String("e", "1-1000", "")
 	excludePorts := flag.String("exclude-ports", "", "")
+	dataDir := flag.String("clients-file", constant.DataDirectory, "")
+	csrFileName := flag.String("csr-filename", DefaultCSRFileName, "")
```
nikita-vanyasin, line 128: Documentation for newly added flags is absent. Please update serverHelp.

constant/const_windows.go
```diff
@@ -0,0 +1,7 @@
+//+build windows
+
+package constant
+
+const (
```
nikita-vanyasin, line 5: Variable value for 'windows' build tag points to non-windows directory (/var/lib/rportd).
Cagent uses ProgramData : https://github.com/cloudradar-monitoring/cagent/blob/master/config.go#L313
I guess rportd should use similar path, though it's better to ask Thorsten.

constant/const.go
```diff
@@ -0,0 +1,7 @@
+//+build !windows
+
+package constant
```
nikita-vanyasin, line 3: I don't really like the idea of having a package with such a common name. I think the constant values should be placed as close to the place where it's used as possible.
Also, it's very probable that in the future we will allow users to change this value via config or CLI argument. (e.g. users can change this value for cagent project).
Therefore it's just a default value for configuration and better to place it near Config type definition.

server/csr/cleanup.go
```diff
@@ -0,0 +1,31 @@
+// Package contains everything related to Client Session Repository (CSR).
+package csr
```
nikita-vanyasin, line 2: I'm not sure 'csr' is a good name for this package because some of defined types are not really related to 'session repository'. e.g. Tunnel type actually implements connection proxying which is not related to a tasks related to a storing any values into persistent or cache storage.
It happened because the Client Session and Tunnel types acting not only as data-object (DTO, which is very good for 'repository' package) but as objects implementing some business logic (creating new tunnels, establishing proxy etc.). But I'm not sure separation of this entities now will bring any help for future development. 
Do we really need that new package? maybe it's enough to create a new package that will know how to store/retrieve data from different storages ?

nikita-vanyasin (changes requested):
Seems good to me except some remarks regarding package naming (see comments).

Also, please fix lint warnings. You can run linter locally using `make docker-golangci-lint` without pushing changes.

server/csr/cleanup.go
m-terel: Agree that after moving other files 'csr' is not an accurate naming for it. What about renaming it to "sessions"?

nikita-vanyasin: :+1:

server/const_windows.go
```diff
@@ -1,8 +1,9 @@
 //+build windows
 
-package main
+package chserver
 
 // Contains constants applicable only to windows OS.
 const (
+	DirectorySeperator   = "\\"
```
nikita-vanyasin, line 7: There is os.PathSeparator for that.

nikita-vanyasin (changes requested):

server/const_windows.go
m-terel: thanks:) fixed.

nikita-vanyasin (approved):

PR #30, opened 2020-09-01
DEV-1688: Migrate to GitHub Workflows

m-terel:


.github/workflows/go.yml
```diff
@@ -0,0 +1,25 @@
...
+          version: v1.30
+          args: -c .golangci.yml
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Test
+        run: go test -v ./...
```
breathbath, line 25: Could you add -race flag here?

.github/workflows/go.yml
```diff
@@ -0,0 +1,25 @@
+name: Go
+on: [push]
+jobs:
+  golangci:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
breathbath, line 8: Why not fetch depth 0 too?

.github/workflows/go.yml
```diff
@@ -0,0 +1,25 @@
...
+
+      - name: go mod vendor
```
breathbath, line 10: Can you check if golangci-lint won’t analyze the vendor folder?

breathbath, line 10: Can you check if golangci-lint won’t analyze the vendor folder?

m-terel: It seems by documentation it's skipped by default. Please see `skip-dirs-use-default` https://golangci-lint.run/usage/configuration/

.github/workflows/go.yml
m-terel: Done.

.github/workflows/go.yml
m-terel: Duplicate comment.

.github/workflows/go.yml
m-terel: Sorry, why do we need to fetch all history for all branches and tags here? I thought it should be done just for a release that is specified in https://goreleaser.com/ci/actions/#workflow

breathbath: Doesn’t 0 means skip history?

breathbath: Doesn’t 0 means skip history?

breathbath (approved):

PR #62, opened 2020-10-05
DEV-1707: Add https cert and key config options

matematik7:
Add `cert_file` and `key_file` config options for api with https.

### DIFF
--- a/share/http_server.go
+++ b/share/http_server.go
@@ -6,6 +6,17 @@ import (
 	"net/http"
 )
 
+var netListen = net.Listen
+
+type ServerOption func(*HTTPServer)
+
+func WithTLS(certFile, keyFile string) ServerOption {
+	return func(s *HTTPServer) {
+		s.certFile = certFile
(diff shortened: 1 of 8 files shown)


share/http_server.go
```diff
@@ -6,34 +6,57 @@ import (
...
+	for _, o := range options {
+		o(s)
+	}
+
+	return s
 }
 
 func (h *HTTPServer) GoListenAndServe(addr string, handler http.Handler) error {
-	l, err := net.Listen("tcp", addr)
+	l, err := netListen("tcp", addr)
```
m-terel, line 47: Please tell why is it moved to a global var?

m-terel (commented):
Looks good. The only thing please update the `api-doc.yml`.

share/http_server.go
matematik7: I was trying to see if I can add tests for this method easily, but then gave up. I reverted this change.

m-terel (approved):

PR #65, opened 2020-10-08
DEV-1754: Add force query param to client delete endpoint.

matematik7:
Add force query param to client delete endpoint.
This will disconnect active sessions and remove disconnected ones for the client being deleted.


server/api.go
```diff
@@ -579,22 +579,49 @@ func (al *APIListener) handleDeleteClient(w http.ResponseWriter, req *http.Reque
...
+	if !force && len(activeSessions) > 0 {
 		al.jsonErrorResponseWithErrCode(w, http.StatusConflict, ErrCodeClientHasSession, fmt.Sprintf("Client expected to have no active session(s), got %d.", len(activeSessions)))
 		return
 	}
+	for _, s := range activeSessions {
+		if err := s.Close(); err != nil {
+			al.jsonErrorResponse(w, http.StatusInternalServerError, err)
+			return
+		}
+		if err := al.sessionService.ForceDelete(s); err != nil {
```
m-terel, line 608: IMO it would be better to have all logic to delete a session in session service. I mean`ForceDelete` could contain all the needed logic:
- if the session is active then it closes the ssh connection, terminates tunnels;
- it removes the session from the cache.

IMO the code would be more clear and better structured when API handler just reads the request, performs some action and returns the response. And all logic related to session deletion is in session service and it can be covered with tests separately. 

So this handler would just get all active&disconnected sessions and then invoke `al.sessionService.ForceDelete(s)` for each of them. What do you think?

server/api_test.go
```diff
@@ -577,6 +593,29 @@ func TestHandleDeleteClient(t *testing.T) {
...
+			sessions:        []*sessions.ClientSession{s1},
+			clientAuthWrite: true,
+			clientID:        cl1.ID,
+			force:           true,
+			wantStatusCode:  http.StatusNoContent,
+			wantClients:     []*clients.Client{cl2, cl3},
+			wantClosedConn:  true,
+		},
+		{
+			descr:           "auth file, client has disconnected session, force",
```
m-terel, line 609: Would be good to add a test case with an invalid force param.

m-terel (commented):
Also please update the swagger api-doc.

server/api.go
```diff
@@ -579,22 +579,49 @@ func (al *APIListener) handleDeleteClient(w http.ResponseWriter, req *http.Reque
 		return
 	}
 
+	force := false
+	forceStr := req.URL.Query().Get("force")
+	if forceStr != "" {
+		var err error
+		force, err = strconv.ParseBool(forceStr)
+		if err != nil {
+			al.jsonErrorResponse(w, http.StatusBadRequest, al.FormatError("invalid force param %v", forceStr))
```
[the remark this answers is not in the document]
matematik7: Sure, usage throughout code base is somewhat inconsistent, so I wasn't sure what to use.

I think parsing is nicer, since it handles different cases, 0 or 1, t or f. It just makes api a bit easier to use.

server/api.go
```diff
@@ -579,22 +579,49 @@ func (al *APIListener) handleDeleteClient(w http.ResponseWriter, req *http.Reque
...
 	}
 
 	activeSessions := al.sessionService.GetByClientID(clientID, true)
-	if len(activeSessions) > 0 {
+	if !force && len(activeSessions) > 0 {
 		al.jsonErrorResponseWithErrCode(w, http.StatusConflict, ErrCodeClientHasSession, fmt.Sprintf("Client expected to have no active session(s), got %d.", len(activeSessions)))
 		return
 	}
+	for _, s := range activeSessions {
+		if err := s.Close(); err != nil {
```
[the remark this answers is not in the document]
matematik7: The tunnels are closed automatically when ssh connection is closed, because the context gets canceled.
I've added a comment about that in `ClientSession.Close`

server/api.go
matematik7: Done

server/api_test.go
matematik7: Done

matematik7:
@m-terel Addressed your comments and added swagger api-doc.

server/api.go
```diff
@@ -579,23 +579,42 @@ func (al *APIListener) handleDeleteClient(w http.ResponseWriter, req *http.Reque
...
 	}
 
 	disconnectedSessions := al.sessionService.GetByClientID(clientID, false)
-	if len(disconnectedSessions) > 0 {
+	if !force && len(disconnectedSessions) > 0 {
 		al.jsonErrorResponseWithErrCode(w, http.StatusConflict, ErrCodeClientHasSession, fmt.Sprintf("Client expected to have no disconnected session(s), got %d.", len(disconnectedSessions)))
 		return
 	}
 
+	allSessions := al.sessionService.GetAllByClientID(clientID)
```
m-terel, line 610: It seems two separate previous calls to get active and disconnected sessions are redundant now. Now we can just return `Client expected to have no session(s), got %d.`. If you agree, please remove them and `func (s *ClientSessionRepository) GetByClientID(clientID string, active bool) []*ClientSession` as well.

matematik7: @m-terel Sure, seems reasonable. Done.

m-terel (approved):

PR #66, opened 2020-10-08
Change redirect 404 to rewrite

matematik7:
Change redirect 404 to rewrite.
When I access path like /dashboard/6, I should be served `index.html` not redirect to it, so that I can refresh the page and stay on it.


m-terel (commented):
A new impl is similar to the previous. Please tell, why wasn't it just changed instead of replacing it with a new file? It just would be easier to review the diff.

matematik7:
@m-terel I moved the file and changed the name since it's not a redirect anymore but a rewrite of url. Usually git would handle movement of file just fine if there are enough similarities, but seems there were too many changes.
I don't think it's a big deal, since there are only a few lines that are actually the same as before. There are quiet a few changes since now it also doesn't propagate header changes until status code is sent.

server/api/middleware/rewrite404_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRewrite404(t *testing.T) {
+	testCases := []struct {
+		Path string
+	}{
+		{
+			Path: "/testpath",
```
m-terel, line 18: nit: maybe a better naming for clarity would be good, `"testpath"->"not_found_path"`?

m-terel (approved):
I see. They still have a lot in common. If the file was renamed, but it just the git shows it as a new file, then it's ok. Good to go.

server/api/middleware/rewrite404_test.go
matematik7: Good idea. Done.

m-terel (approved):
I see. They still have a lot in common. If the file was renamed, but it just the git shows it as a new file, then it's ok. Good to go.

PR #67, opened 2020-10-09
DEV-1746: Add access log for api

matematik7:


server/api/middleware/accesslog.go
```diff
@@ -0,0 +1,80 @@
...
+	"net/http"
+	"sync"
+	"time"
+
+	chshare "github.com/cloudradar-monitoring/rport/share"
+)
+
+var timeNow = time.Now
+
+type AccessLogMiddleware struct {
```
m-terel, line 16: Why just don't use https://godoc.org/github.com/gorilla/handlers#LoggingHandler ?

server/api_listener.go
```diff
@@ -81,6 +84,14 @@ func NewAPIListener(
 		clientProvider:    clientProvider,
 	}
 
+	if config.API.AccessLogFile != "" {
+		accessLogFile, err := os.OpenFile(config.API.AccessLogFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
m-terel, line 88: Just to clarify. If dir `/var/log/rport/` for logs does not exist, it fails. Maybe we should try to create the dir?

rportd.example.conf
```diff
@@ -218,6 +218,10 @@
   # cert_file = "/var/lib/rport/server.crt"
   # key_file = "/var/lib/rport/server.key"
 
+  ## Specifies file for API access logs. Logs will be written in Combined Log Format.
```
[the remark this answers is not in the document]
matematik7: Done

server/api/middleware/accesslog.go
matematik7: Honestly, I didn't even go looking for a library for something this simple.

But also if we ever decide we need to log the user, then we can integrate this with our auth, we can't do that with gorilla.

server/api_listener.go
matematik7: No we don't create the dir, I don't think it's expected.
And since rportd shouldn't be running as root it wouldn't have permission to create the dir anyway.

server/api/middleware/accesslog.go
m-terel: If it would be so simple then it can be done like in one hour. But it requires much more time. So it's not so simple. Moreover, it's additional 150 lines of code that should be further maintained.

Please tell why we can't do that with gorilla? I've tried to use `CombinedLoggingHandler` and just set `r.URL.User = url.User(username)` in our auth middleware and it seems it works just fine.
 
Since we are using gorilla and there is already logging middleware that is widely used why just don't use it if it does everything that we need? Is it worth to keep our own implementation that will require to maintain it in the future?

m-terel: Also, let's maybe log the user as well? Please tell what is the reason not to do that now?

m-terel: My suggestion is to use the existing gorilla implementation if it's everything that we need to avoid maintaining extra code. But if you prefer to have your own implementation, then I wouldn't insist on using gorilla. But please add the logging of the user. IMO it should be useful to have it there.

matematik7: Alright, if we can set the user, then I guess using a well established library is better. Thanks for pointing this out.

Since our auth middleware is lower in the stack then logging middleware, we can't access the user from logging middleware. To be able to log the user I would have to refactor how authentication is handled. I asked Thorsten and he told me that logging the user is not that important for now and not to spend more time on it.

m-terel (approved):

PR #69, opened 2020-10-13
DEV-1646: Feature to run remote command, refactoring

m-terel:


client/exec_nix.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+// now is used to stub time.Now in tests
+var now = func() time.Time {
+	return time.Now()
+}
+
+func (c *Client) HandleRunCmdRequest(ctx context.Context, reqPayload []byte) (*comm.RunCmdResponse, error) {
+	// do not accept a new request when the previous is not finished yet
+	curPID := c.getCurCmdPID()
```
matematik7, line 24: I see a potential for a race condition if you get another cmd request between this being read and the pid being set below.
It would be better to have the mutex locked after read until you write the new value.

client/exec_nix_test.go
```diff
@@ -0,0 +1,139 @@
...
+`
+	// when
+	res, err := c.HandleRunCmdRequest(context.Background(), []byte(jobToRunJSON))
+	// mimic real behavior and wait until background task finishes
+	time.Sleep(100 * time.Millisecond)
+
+	// then
+
+	// check returned result
+	assert.NoError(err)
```
matematik7, line 85: From what I've seen in other tests we usually use `require` for checking error that should not happen in tests and `assert` for testing test results. Let's keep it consistent.

share/comm/ssh.go
```diff
@@ -37,26 +37,30 @@ func ReplySuccessJSON(log *chshare.Logger, req *ssh.Request, resp interface{}) {
...
 }
 
-// HandleSSHRequestJSON sends a given request, parses the returned response and stores the result in a given destination value.
-// Both request and response are expected to be JSON. Returns an error on a failure response or if an error happen.
-func HandleSSHRequestJSON(conn ssh.Conn, reqType string, req, respDest interface{}) error {
+// SendRequestAndGetResponse sends a given request, parses a returned response and stores a success result in a given destination value.
+// Returns (nil, err) if an internal error occurred (i.e. a request encoding error, an error on sending a request).
+// Returns (failureMsg, nil) on a failure response or a response decoding error.
+// Both request and response are expected to be JSON.
+func SendRequestAndGetResponse(conn ssh.Conn, reqType string, req, successRespDest interface{}) (failureMsg *string, err error) {
```
matematik7, line 44: I've never seen two different error return types.
You should only have error as return type. If you need to differentiate errors in the caller for some reason, you can assert on the custom error type.
https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go

client/exec_nix.go
m-terel: @matematik7 `HandleRunCmdRequest` is run sequentially and not in a separate goroutine. That's why it should not have any race condition.

client/exec_nix.go
```diff
@@ -0,0 +1,104 @@
...
+		var status string
+		select {
+		case err := <-done:
+			if err != nil {
+				status = models.JobStatusFailed
+				c.Errorf("failed to run command[jid=%q,pid=%d]:\ncmd:\n%s\nerr: %s", job.JID, res.Pid, job.Command, err)
+			} else {
+				status = models.JobStatusFinished
+			}
+		case <-time.After(job.Timeout):
```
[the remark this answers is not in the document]
m-terel: @matematik7 as commented here https://github.com/cloudradar-monitoring/rport/blob/5aeed27d668474157dcfe4928a6e5859f03dd817/client/exec_nix.go#L61
we should just stop the observing, but cmd execution should proceed.  That's why `context.WithTimeout` wasn't used.

server/api.go
```diff
@@ -638,10 +662,161 @@ func (al *APIListener) allowClientAuthWrite(w http.ResponseWriter) bool {
...
+		al.jsonErrorResponseWithTitle(w, http.StatusBadRequest, fmt.Sprintf("Missing %q route param.", routeParamSessionID))
+		return
+	}
+
+	reqBody := struct {
+		Command    string `json:"command"`
+		TimeoutSec int    `json:"timeout_sec"`
+	}{}
+	err := json.NewDecoder(req.Body).Decode(&reqBody)
+	if err == io.EOF {
```
matematik7, line 690: Why handle this separately? If it's an error then a generic error response should be fine.

server/server.go
```diff
@@ -26,14 +29,23 @@ type Server struct {
 	config         *Config
 }
 
+// sharedOptions data shared between API Listener and Client Listener
+type sharedOptions struct {
```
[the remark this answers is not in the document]
m-terel: Done.

server/api/jobs/file_test.go
```diff
@@ -0,0 +1,448 @@
...
+		runningJobs          []*models.Job
+		sid                  string
+		jid                  string
+		fReturnExist         bool
+		fReturnExistErr      error
+		fReturnReadFileErr   error
+		fSetReadFileDestFunc func(dest interface{})
+
+		wantExistPath   bool
+		wantReadFile    bool
```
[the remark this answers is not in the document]
matematik7: I usually prefer to have a fake in memory files implementation, that behaves like real one, so you don't have to test function arguments and calls.
But I guess this is fine as well.

PR #70, opened 2020-10-13
DEV-1711: Prevent rport(d) running as root.

matematik7:
Prevent rport(d) running as root unless allow root is set.

### DIFF
--- a/cmd/rport/main.go
+++ b/cmd/rport/main.go
@@ -106,6 +106,9 @@ var clientHelp = `
     Used for filtering clients on the server.
     Can be used multiple times. (e.g --tag "foobaz" --tag "bingo")
 
+    --allow-root, An optional arg to allow running rport as root. There is no technical requirement to run the rport
+    client under the root user. Running it as root is an unnecessary security risk.
+
     --service, Manages rport running as a service. Possible commands are "install", "uninstall", "start" and "stop".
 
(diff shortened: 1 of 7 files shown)


cmd/rport/main.go
```diff
@@ -228,6 +233,10 @@ func runMain(cmd *cobra.Command, args []string) {
 		log.Fatal(err)
 	}
 
+	if !config.Client.AllowRoot && os.Geteuid() == 0 {
```
m-terel, line 236: Without a research it's not clear that `os.Geteuid() == 0` means it's run with a root access. Maybe would be good to create a separate func. The func name would tell what it does and it can be reused by `rport` and `rportd`.

rport.example.conf
```diff
@@ -69,6 +69,12 @@ server = "rportd.lan:9000"
 #  '5050'
 #]
 
+## There is no technical requirement to run the rport client under the root user.
+## Running it as root is an unnecessary security risk.
+## Rport exists with an error if started as root unless you explicitly allow it.
```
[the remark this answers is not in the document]
matematik7: Fixed

cmd/rport/main.go
matematik7: Done

m-terel (approved):

PR #72, opened 2020-10-14
DEV-1763: Refactor client config

matematik7:
Refactor client config:
- move all config parsing to `config.ParseAndValidate` from `NewClient`
- add config as direct attribute of `Client`
- `NewClient` does not return error
- add tests

There shouldn't be any changes in functionality in this pr.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -28,7 +28,7 @@ type Client struct {
 	*chshare.Logger
 	connReq *chshare.ConnectionRequest
 
-	Config         *Config
+	config         *Config
 	sshConfig      *ssh.ClientConfig
 	sshConn        ssh.Conn
 	running        bool
@@ -53,18 +53,16 @@ func NewClient(config *Config) *Client {
 
 	client := &Client{
 		Logger:   chshare.NewLogger("client", config.Logging.LogOutput, config.Logging.LogLevel),
-		Config:   config,
+		config:   config,
 		connReq:  connectionReq,
 		running:  true,
 		runningc: make(chan error, 1),
 		cmdExec:  NewCmdExecutor(),
 	}
 
-	user, pass := chshare.ParseAuth(config.Client.Auth)
-
 	client.sshConfig = &ssh.ClientConfig{
-		User:            user,
-		Auth:            []ssh.AuthMethod{ssh.Password(pass)},
+		User:            config.Client.authUser,
+		Auth:            []ssh.AuthMethod{ssh.Password(config.Client.authPass)},
 		ClientVersion:   "SSH-" + chshare.ProtocolVersion + "-client",
 		HostKeyCallback: client.verifyServer,
 		Timeout:         30 * time.Second,
@@ -85,7 +83,7 @@ func (c *Client) Run() error {
 
 func (c *Client) verifyServer(hostname string, remote net.Addr, key ssh.PublicKey) error {
 	got := chshare.FingerprintKey(key)
-	if c.Config.Client.Fingerprint != "" && !strings.HasPrefix(got, c.Config.Client.Fingerprint) {
+	if c.config.Client.Fingerprint != "" && !strings.HasPrefix(got, c.config.Client.Fingerprint) {
 		return fmt.Errorf("Invalid fingerprint (%s)", got)
 	}
 	//overwrite with complete fingerprint
@@ -96,13 +94,13 @@ func (c *Client) verifyServer(hostname string, remote net.Addr, key ssh.PublicKe
 //Start client and does not block
 func (c *Client) Start(ctx context.Context) error {
 	via := ""
-	if c.Config.Client.proxyURL != nil {
-		via = " via " + c.Config.Client.proxyURL.String()
+	if c.config.Client.proxyURL != nil {
+		via = " via " + c.config.Client.proxyURL.String()
 	}
 
-	c.Infof("Connecting to %s%s\n", c.Config.Client.Server, via)
+	c.Infof("Connecting to %s%s\n", c.config.Client.Server, via)
 	//optional keepalive loop
-	if c.Config.Connection.KeepAlive > 0 {
+	if c.config.Connection.KeepAlive > 0 {
 		go c.keepAliveLoop()
 	}
 	//connection loop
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -29,77 +28,36 @@ type Client struct {
...
-	proxyURL          *url.URL
-	server            string
-	serverFingerprint string
-	running           bool
-	runningc          chan error
-	connStats         chshare.ConnStats
-	cmdExec           CmdExecutor
-	curCmdPID         *int
-	curCmdPIDMutex    sync.Mutex
+	Config         *Config
```
m-terel, line 31: Is it public intentionally?

client/config.go
```diff
@@ -33,6 +35,9 @@ type ClientConfig struct {
 	Name        string   `mapstructure:"name"`
 	Tags        []string `mapstructure:"tags"`
 	Remotes     []string `mapstructure:"remotes"`
+
+	proxyURL *url.URL
+	remotes  []*chshare.Remote
```
m-terel, line 40: Maybe would be good to add an auth user here to avoid parsing it in `NewClient`? And do the parsing in `ParseAndValidate`
https://github.com/cloudradar-monitoring/rport/blob/975967d7be81bc11491506ae2a6d5eda1f0bc81d/client/client.go#L63

client/config.go
```diff
@@ -63,6 +87,56 @@ func (c *Config) ParseAndValidate() error {
 	return nil
 }
 
+func (c *Config) parseServerURL() error {
+	if c.Client.Server == "" {
+		return fmt.Errorf("Server address is required. See --help")
```
m-terel, line 92: Please fix the capitalized first letter. In other places as well.

`fmt.Errorf`->`fmt.Error`

client/config_test.go
```diff
@@ -57,10 +61,214 @@ func TestConfigParseAndValidate(t *testing.T) {
...
+			MaxRetryInterval:         time.Second,
+			ExpectedMaxRetryInterval: time.Second,
+		}, {
+			Name:                     "set max retry interval",
+			MaxRetryInterval:         time.Minute,
+			ExpectedMaxRetryInterval: time.Minute,
+		}, {
+			Name:                     "default",
+			MaxRetryInterval:         time.Duration(0),
+			ExpectedMaxRetryInterval: 5 * time.Minute,
```
m-terel, line 147: Maybe add one more test where `0 < MaxRetryInterval < time.Second`

client/config.go
```diff
@@ -63,6 +87,56 @@ func (c *Config) ParseAndValidate() error {
...
+		c.Client.proxyURL = proxyURL
+	}
+	return nil
+}
+
+func (c *Config) parseRemotes() error {
+	for _, s := range c.Client.Remotes {
+		r, err := chshare.DecodeRemote(s)
+		if err != nil {
+			return fmt.Errorf("Failed to decode remote '%s': %s", s, err)
```
m-terel, line 133: nit: capital 1st letter and `'%s'->%q`

m-terel (commented):
Good job on refactoring. We should continue to refactor parts that are written not well and cover code with tests.

client/client.go
```diff
@@ -29,77 +28,36 @@ type Client struct {
...
-	u.Scheme = strings.Replace(u.Scheme, "http", "ws", 1)
+func NewClient(config *Config) *Client {
 	connectionReq := &chshare.ConnectionRequest{
 		Version: chshare.BuildVersion,
 		ID:      config.Client.ID,
 		Name:    config.Client.Name,
 		Tags:    config.Client.Tags,
+		Remotes: config.Client.remotes,
 	}
 	connectionReq.OS, _ = chshare.Uname()
```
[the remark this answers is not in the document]
matematik7: I will refactor connection request in next pr anyway, so I won't do this here.

client/client.go
```diff
@@ -29,77 +28,36 @@ type Client struct {
...
-	running           bool
-	runningc          chan error
-	connStats         chshare.ConnStats
-	cmdExec           CmdExecutor
-	curCmdPID         *int
-	curCmdPIDMutex    sync.Mutex
+	Config         *Config
+	sshConfig      *ssh.ClientConfig
+	sshConn        ssh.Conn
+	running        bool
```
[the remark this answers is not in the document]
matematik7: While I agree the running and runningc could be done nicer with `Context`, I don't think it's critical and it's existing code that's not part of this config refactor.

client/client.go
```diff
@@ -29,77 +28,36 @@ type Client struct {
...
+	runningc       chan error
```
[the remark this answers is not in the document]
matematik7: See above.

client/client.go
```diff
@@ -29,77 +28,36 @@ type Client struct {
...
-
-	if p := config.Client.Proxy; p != "" {
-		client.proxyURL, err = url.Parse(p)
-		if err != nil {
-			return nil, fmt.Errorf("Invalid proxy URL (%s)", err)
-		}
+		Logger:   chshare.NewLogger("client", config.Logging.LogOutput, config.Logging.LogLevel),
+		Config:   config,
+		connReq:  connectionReq,
+		running:  true,
```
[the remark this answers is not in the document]
matematik7: See above.

client/client.go
matematik7: Changed.

client/config.go
matematik7: Done.

client/config.go
matematik7: Done.

client/config_test.go
matematik7: Done.

client/config.go
matematik7: Done.

client/client.go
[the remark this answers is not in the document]
m-terel: `Client.Close()` wasn't used before. Now in `service.go` it's used, so there is a potential race condition that I mentioned. Maybe would be good to add at least TODO to fix it to avoid forgetting it?

m-terel (approved):

matematik7:
I don't really like TODOs in code, they tend to get out of date or forgotten. If something needs to be done it's better to create a ticket.

PR #77, opened 2020-10-20
DEV-1797: GitHub Workflows: auto-deploy rport clients (Windows)

m-terel:


.github/workflows/deploy.yml
```diff
@@ -48,6 +48,7 @@ jobs:
 
   unix-clients:
     name: Unix Clients
+    needs: server
```
matematik7, line 51: Why should clients wait for server, I would think parallel deploy should be fine?

.github/workflows/deploy.yml
```diff
@@ -75,5 +76,35 @@ jobs:
...
+  win-clients:
+    name: Windows Clients
+    needs: server
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build
+        run: go build -v -o ./ ./cmd/rport/...
```
matematik7, line 91: If we don't set the `BuildVersion` here, it will be hard to figure out what version is deployed.

.github/workflows/deploy.yml
```diff
@@ -75,5 +76,35 @@ jobs:
...
+          remote: "/home/uploader/rport/rport.exe"
+
+      - name: Restart clients
+        uses: garygrossgarten/github-action-ssh@release
+        with:
+          host: ${{ secrets.CLIENTS_HOST }}
+          username: ${{ secrets.CLIENTS_USERNAME }}
+          port: ${{ secrets.CLIENTS_PORT }}
+          privateKey: ${{ secrets.CLIENTS_KEY }}
+          command: sudo salt "test-win*" state.apply rport.win.dev-client
```
matematik7, line 110: I've never used salt before. Why does this only need `state.apply` while linux also does `service.restart`?

.github/workflows/deploy.yml
m-terel: It was done in case some breaking changes were introduced by a commit that requires a new version for both server and client. The wait would resolve a case when the server failed to deploy but clients were deployed successfully. In general, this should not happen, but it's possible. But maybe you are right, maybe we should not wait.

.github/workflows/deploy.yml
m-terel: Sure, I just have trouble to set it as before in a format `{date-time}-sha:{github-master-head-sha}`. Any suggestion on how it can be quickly done?

.github/workflows/deploy.yml
m-terel: You can check out salt scripts that are on `hv3.cloudradar.info` in `/srv/salt/rport/`. I'm kind of new to salt as well, but while I was creating these scripts it turned out that for unix it was ok just to replace the binary and then restart the service. It allows to minimize the downtime. But for windows it fails if you try to replace the binary while the rport service is running. That's why in salt script for windows at first I stop the rport service and then I replace the binary. If the rport service is stopped and the salt state is applied it will start the rport client with new changes without any need to restart the service.

.github/workflows/deploy.yml
matematik7: @m-terel Honestly, I have no clue about windows command line. I would build on linux with `GOOS=windows`

m-terel: @matematik7 thanks, will use your suggestion

PR #78, opened 2020-10-20
DEV-1644: Client auth refactoring

matematik7:
- Move client provider to `clients` package, it shouldn't be in server.
- Make cache use provider, so I'll be able to add Add and Delete methods to provider.
- Do not use client cache in client listener for ssh auth caching since it's completely different purpose and not even necessary.


server/clients/cache.go
```diff
@@ -7,17 +7,25 @@ import (
...
+	provider Provider
+	clients  map[string]*Client
+	mu       sync.RWMutex
 }
 
 // NewClientCache returns a thread-safe cache with ID as a key populated with given clients.
-func NewClientCache(initClients []*Client) *ClientCache {
-	m := make(map[string]*Client, len(initClients))
-	for _, v := range initClients {
+func NewClientCache(provider Provider) (*ClientCache, error) {
```
m-terel, line 16: With these changes this `ClientCache` is a wrapper over a provider.  Would be good to use a better naming. Maybe`CachedProvider`?

This `CachedProvider` is also a provider, but it just adds caching to the existing provider. My suggestion is to make it to implement the `Provider` interface and in `Server` use `Provider` interface instead of an implementation. It would add an abstraction and flexibility, would make code less coupled. So we would decide what provider we want when `Server` is created. For example, at first we can use cached DBProvider and then if we decide that we don't need caching for DB then this switch would be very easy, just one line of code.

matematik7: @m-terel Seems like a good idea, but for now client cache and provider still have different interfaces.
I'll work towards this goal in future prs when I add `Add` and `Delete` methods to provider.

m-terel (approved):

PR #79, opened 2020-10-20
DEV-1646: Remote command execution improvements

m-terel:
- Use prefixes in job file name to be able to distinguish the command status without opening the file.


client/exec_nix.go
```diff
@@ -80,13 +80,17 @@ func (c *Client) HandleRunCmdRequest(ctx context.Context, reqPayload []byte) (*c
...
 		job.FinishedAt = &now
-		job.Status = status
 		job.PID = res.Pid
 		job.StartedAt = startedAt
 		job.Result = &models.JobResult{
 			StdOut: stdOut.String(),
 			StdErr: stdErr.String(),
 		}
+		// if stderr is not empty set the status to failed
+		if job.Result.StdErr != "" {
```
matematik7, line 90: This seems wrong, command can write to stderr without failing. Failed status should be determined based on exit code.

server/api/jobs/file.go
```diff
@@ -138,17 +160,22 @@ func convertToJobSummary(file os.FileInfo) *models.JobSummary {
 	}
 
 	fileName := file.Name()
-	jid := fileName[:len(fileName)-len(jobFileExtension)]
+	prefix := fileName[:prefixLength]
+	status := prefixStatusMapping[prefix]
```
matematik7, line 164: I'm not sure I like this prefix status, are you sure this is needed? Why not just open the file and read the status? Seems like premature optimization to me.

client/exec_nix.go
m-terel: Agreed to keep the existing logic since cmd can write to stderr without failing.

server/api/jobs/file.go
m-terel: Agreed in chat it's worth to get short info about the all cmds based on the file metadata instead of loading all files that can be heavy.

matematik7 (approved):

PR #80, opened 2020-10-21
DEV-1801: Changes around verbose and updates in config

m-terel:


rportd.example.conf
```diff
@@ -210,7 +210,8 @@
   ## If specified, rportd will serve static files from this directory on the same API address.
   ## This is the place where the frontend files (html/js) go.
   ## Learn how to install the frontend https://github.com/cloudradar-monitoring/rport/blob/master/docs/frontend.md
-  doc_root = "/var/lib/rport/docroot"
+  #doc_root = "C:\Program Files\rport\docroot"
```
matematik7, line 213: Other entries have space after `#`.

share/logger.go
```diff
@@ -15,6 +16,9 @@ const (
 )
 
 func ParseLogLevel(str string) (LogLevel, error) {
+	if str == "" {
+		return LogLevelError, errors.New("log level not specified")
```
matematik7, line 20: Not sure why this is needed since we have a default log level?

matematik7 (approved):

rportd.example.conf
m-terel: Actually, not all of them. It's mixed, some with space, some without.

matematik7: Well, I don't really care one way or another. Do you mind making it consistent?

share/logger.go
m-terel: Removed.

rportd.example.conf
m-terel: done

matematik7 (approved):

PR #81, opened 2020-10-21
DEV-1644: Remove no client auth support

matematik7:
According to Thorsten using rport without client auth should not be supported, so let's remove it (and it didn't work anyway :D).


matematik7:
@m-terel No clients could connect if no client auth was specified, since you need to set `NoClientAuth` for ssh server to accept all clients.

m-terel:
I see :) thanks

m-terel (approved):

PR #82, opened 2020-10-21
DEV-1644: Use session lookup instead of client locking.

matematik7:
Clients were locked to specific session id while session was connected or saved.
Instead of locking we can just check session repo to see if session with different id exists.
This is much simpler and removes unnecessary functionality from client cache.


server/client_listener.go
```diff
@@ -213,28 +213,12 @@ func (cl *ClientListener) handleWebsocket(w http.ResponseWriter, req *http.Reque
 		}
 	}
 
-	// lock client credentials if needed
-	locked, err := cl.lockClientIfNeeded(client, sid)
-	if err != nil {
-		failed(err)
+	// check if client is already used by another session
+	if !cl.config.Server.AuthMultiuseCreds && cl.isClientInUse(client.ID, sid) {
```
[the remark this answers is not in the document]
matematik7: You're right, added mutex while querying session service and adding the session.

server/client_listener.go
```diff
@@ -439,3 +373,48 @@ func (cl *ClientListener) handleSSHChannels(clientLog *chshare.Logger, chans <-c
...
+func (cl *ClientListener) isClientInUse(clientID, sessionID string) bool {
+	for _, s := range cl.sessionService.GetAllByClientID(clientID) {
+		if s.ID != sessionID {
+			return true
+		}
+	}
+	return false
+}
+
+func (cl *ClientListener) startSession(ctx context.Context, clog *chshare.Logger, connRequest *chshare.ConnectionRequest, sshConn ssh.Conn, client *clients.Client, sid string) (*sessions.ClientSession, error) {
```
m-terel, line 387: It seems this logic belongs to `sessionService.StartClientSession`. My suggestion is to move it there instead of having this func here. And the mutex should be in session service, since there are other read/write funcs, so they all should be guarded by it to avoid race conditions.

m-terel: Also, would be great if you can add few tests for `startSession`.

m-terel:
Good job on simplifying the logic👍

server/client_listener.go
matematik7: @m-terel Done.

server/client_listener.go
```diff
@@ -193,54 +193,13 @@ func (cl *ClientListener) handleWebsocket(w http.ResponseWriter, req *http.Reque
...
-			}
-		}()
-	}
-
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	clientSession, err := cl.sessionService.StartClientSession(ctx, sid, sshConn, connRequest, client, clog)
+	clientSession, err := cl.sessionService.StartClientSession(
+		ctx, client.ID, sid, sshConn, cl.config.Server.AuthMultiuseCreds, connRequest, clog)
```
m-terel, line 200: `client.ID` it seems it can lead to `nil pointer` in case somebody simultaneously deletes the client. Sure, it's a low probability, my suggestion to add an `if` and return an error since now we require client auth. Or drop the `client` and just use `sshConn.User()` as clientID.

server/session_service.go
```diff
@@ -16,6 +17,8 @@ import (
 type SessionService struct {
 	repo            *sessions.ClientSessionRepository
 	portDistributor *ports.PortDistributor
+
+	mtx sync.Mutex
```
m-terel, line 21: Since `Terminate` updates `session.Disconnected` and `ForceDelete` and read funcs read this value wouldn't it lead to a race condition? My suggestion is to use a read-write lock and protect all funcs.

nit: to be consistent maybe rename to `mu`?

server/client_listener.go
matematik7: Changed to using clientID only.

server/session_service.go
matematik7: @m-terel Done.

m-terel: @matematik7 sorry, shouldn't we use a read-write lock and lock all read operations as well? Without it it seems there can be a race condition when `session.Disconnected` is set in `Terminate` and simultaneously this value is read in https://github.com/cloudradar-monitoring/rport/blob/6a8fd247084cef39660f3f6cc2f20ceee9ce412b/server/sessions/session.go#L55-L56 that is used by `Get` funcs.

matematik7: @m-terel If we want to make sure session reading and writing is consistent throughout the whole app that's a much bigger task and I'm not tackling it in the scope of this pr. I've added locks for this service.

m-terel (approved):

PR #83, opened 2020-10-22
DEV-1646: Remote cmd execution improvements

m-terel:
- Use `timeout_sec` in both the `POST` request object and the job response to avoid confusion.
- Fail the `POST` request if its body contains unknown params.


server/api.go
```diff
@@ -664,7 +664,9 @@ func (al *APIListener) handlePostCommand(w http.ResponseWriter, req *http.Reques
 		Command    string `json:"command"`
 		TimeoutSec int    `json:"timeout_sec"`
 	}{}
-	err := json.NewDecoder(req.Body).Decode(&reqBody)
+	dec := json.NewDecoder(req.Body)
+	dec.DisallowUnknownFields()
```
matematik7, line 668: I think this setting for json decoder should be consistent across all api endpoints.

matematik7 (approved):

server/api.go
m-terel: It should. It can be done in a separate PR.

PR #84, opened 2020-10-22
DEV-1644: Refactor file and cached provider

matematik7:
Refactor file and cached provider to both have Provider interface and use that in server.
Also change file to write through instead of writing every n seconds.


server/clients/provider.go
```diff
@@ -1,7 +1,13 @@
 package clients
 
 type Provider interface {
+	// Get returns client from provider or nil
+	Get(id string) (*Client, error)
```
m-terel, line 5: nit: maybe would be better to use `GetByID` `DeleteByID`? To avoid the situation when a new method by another search param will be introduced.

server/clients/provider.go
```diff
@@ -10,10 +16,46 @@ type mockProvider struct {
...
+	existing, err := p.Get(id)
+	if err != nil {
+		return false, err
+	}
+	if existing != nil {
+		return false, nil
+	}
+	p.clients = append(p.clients, client)
+	return true, nil
+}
```
m-terel, line 46: nit: newline

server/clients/provider.go
```diff
@@ -10,10 +16,46 @@ type mockProvider struct {
...
+func (p *mockProvider) Delete(id string) error {
```
m-terel, line 47: This mock impl seems to be not so simple and therefore can have some race conditions (our tests are run with `race` so there is a potential test failure). Wouldn't be easier just to use a thread-safe map?

server/clients/file.go
```diff
@@ -4,54 +4,115 @@ import (
...
-type FileClients struct {
+// FileProvider is not thread save so should be surrounded by CachedProvider.
+type FileProvider struct {
 	log      *chshare.Logger
 	fileName string
 }
 
-func NewFileClients(log *chshare.Logger, fileName string) *FileClients {
-	return &FileClients{
+func NewFileProvider(log *chshare.Logger, fileName string) *FileProvider {
```
m-terel, line 18: It seems `log *chshare.Logger` is not used anymore.

server/clients/file.go
```diff
@@ -4,54 +4,115 @@ import (
...
+	idPswdPairs, err := c.load()
 	if err != nil {
-		return nil, fmt.Errorf("failed to read rport clients auth file %q: %s", c.fileName, err)
+		return nil, fmt.Errorf("failed to decode rport clients auth file: %v", err)
+	}
+
+	var res []*Client
+	for id, pswd := range idPswdPairs {
+		if id == "" || pswd == "" {
+			return nil, fmt.Errorf("empty client ID or password is not allowed")
```
m-terel, line 35: nit: `Errorf` -> `errors.New`

server/clients/single.go
```diff
@@ -1,33 +1,48 @@
...
+		},
 	}
 }
 
 // GetAll returns a list with a single client credentials.
-func (c *SingleClient) GetAll() ([]*Client, error) {
+func (c *SingleProvider) GetAll() ([]*Client, error) {
 	c.log.Infof("Parsing single client auth credentials...")
 
-	client := &Client{}
```
m-terel, line 25: Please remove the logging above. And logger since it wouldn't be used anymore.

m-terel (commented):
Also, it seems `README` file should be updated since now it's required to specify the client auth method.

server/clients/cache.go
```diff
@@ -4,15 +4,15 @@ import (
 	"sync"
 )
 
-// ClientCache is a thread-safe in-memory cache.
-type ClientCache struct {
+// CachedProvider is a thread-safe in-memory cache.
```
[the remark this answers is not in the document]
matematik7: Updated the doc.

server/clients/cache.go
```diff
@@ -21,69 +21,55 @@ func NewClientCache(provider Provider) (*ClientCache, error) {
...
-	c.clients[key] = client
+	return res, nil
 }
 
-// Add returns true if a new client by a given key was added successfully.
-// Returns false if it already contains a client with such key.
-func (c *ClientCache) Add(key string, client *Client) bool {
+// Add returns true if a new client by a given id was added successfully.
+// Returns false if it already contains a client with such id.
+func (c *CachedProvider) Add(id string, client *Client) (bool, error) {
```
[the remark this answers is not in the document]
matematik7: Done.

server/clients/file.go
```diff
@@ -4,54 +4,115 @@ import (
...
+		}
+		res = append(res, &Client{ID: id, Password: pswd})
 	}
-	c.log.Infof("Parsing rport clients data...")
 
-	res, err := decodeClients(b)
+	return res, nil
+}
+
+func (c *FileProvider) Get(id string) (*Client, error) {
```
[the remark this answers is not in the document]
matematik7: It's a really simple implementation since most of the logic is in `load` anyway, so I think it's better to just make it work.

server/clients/file.go
```diff
@@ -4,54 +4,115 @@ import (
...
-	for id, pswd := range idPswdPairs {
-		if id == "" || pswd == "" {
-			return nil, fmt.Errorf("empty client ID or password is not allowed")
-		}
-		res = append(res, &Client{ID: id, Password: pswd})
+	return idPswdPairs, nil
+}
+
+func (c *FileProvider) save(idPswdPairs map[string]string) error {
+	file, err := os.OpenFile(c.fileName, os.O_RDWR|os.O_TRUNC, os.ModePerm)
```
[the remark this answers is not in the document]
matematik7: There's a comment above that file provider is not thread save and has to be guarded by `CachedProvider` which locks all methods around the underlying provider.

server/clients/provider.go
```diff
@@ -1,7 +1,13 @@
 package clients
 
 type Provider interface {
+	// Get returns client from provider or nil
+	Get(id string) (*Client, error)
 	GetAll() ([]*Client, error)
+	// Add returns true if the client was added and false if it already exists
+	Add(id string, client *Client) (bool, error)
+	Delete(id string) error
```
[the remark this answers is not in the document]
matematik7: Done

server/clients/provider.go
matematik7: I think Get and Delete by themselves sort of imply that they're by primary key, while if you have a method called `GetByX` then it implies a search and returning multiple results.

server/clients/provider.go
matematik7: Done

server/clients/provider.go
matematik7: Changed the implementation to map since it is simpler, but I left it non thread safe and added a comment. Our tests don't run in parallel so this should not be needed, but if it ever is the implementation can be changed.

server/clients/file.go
matematik7: Removed

server/clients/file.go
matematik7: Done

server/clients/single.go
matematik7: Done

README.md
```diff
@@ -88,7 +88,7 @@ tar vxzf - rport -C /usr/local/bin/
 ```
 
 Create an ad hoc tunnel that will forward the port 2222 of node1.example.com to the to local port 22 of client1.local.localdomain.
-`rport node1.example.com:19075 2222:0.0.0.0:22`
+`rport --auth rport:password123 node1.example.com:19075 2222:0.0.0.0:22`
```
m-terel, line 91: Please add the auth option everywhere for `rport` and `rportd` examples (especially in quick start), because without it it would fail.

m-terel (approved):

README.md
matematik7: Done.

m-terel (approved):

PR #85, opened 2020-10-22
DEV-1791-1812: Execute remote command (Windows)

m-terel:


client/exec.go
```diff
@@ -23,3 +31,117 @@ func (e *CmdExecutorImpl) Start(cmd *exec.Cmd) error {
...
+		job.StartedAt = startedAt
+		job.Result = &models.JobResult{
+			StdOut: stdOut.String(),
+			StdErr: stdErr.String(),
+		}
+
+		// send the filled job to the server
+		jobBytes, err := json.Marshal(job)
+		if err != nil {
+			c.Errorf("failed to send command result for [jid=%q,pid=%d]: failed to encode job result: %s", job.JID, res.Pid, err)
```
matematik7, line 136: Perhaps there should be a return here if marhsaling fails, you have nothing to send?

client/exec_win_test.go
```diff
@@ -0,0 +1,54 @@
+//+build windows
```
matematik7, line 1: Not sure how this work, to these tests run only on windows? Do we need a separate step in our CI to run these?

matematik7 (dismissed):

client/exec.go
m-terel: Good point, fixed.

client/exec_win_test.go
m-terel: If we introduce a test with `//+build windows` then I believe so.

Another way, I can rewrite code to have smth like `getShell(inputShell, os string) (string, error)` to avoid having separate `_nix.go` and `_win.go` implementations. But it complicates a logic a little bit, because the logic should be combined for both os. @matematik7 what option do you prefer?

matematik7: I think if we can avoid having different code compiled for different os and instead can just use a bit more logic I think that should be prefered.

m-terel: Agree, updated.

client/exec.go
```diff
@@ -23,3 +32,136 @@ func (e *CmdExecutorImpl) Start(cmd *exec.Cmd) error {
...
+	unixShell: {"-c"},
+	cmdShell:  {"/c"},
+	powerShell: {
+		"-Noninteractive", // Don't present an interactive prompt to the user.
+		"-Command",
+	},
+}
+
+// now is used to stub time.Now in tests
+var now = func() time.Time {
```
matematik7, line 53: This could be simplified: `var now = time.Now`

client/exec_test.go
```diff
@@ -73,3 +86,220 @@ func (e *CmdExecutorMock) Wait(cmd *exec.Cmd) error {
...
+	require.NoError(t, err)
+	assert.Equal(&comm.RunCmdResponse{Pid: wantPID, StartedAt: nowMock}, res)
+
+	// check job result that was sent to server
+	inputRequestName, inputWantReply, inputPayload := connMock.InputSendRequest()
+	assert.Equal(comm.RequestTypeCmdResult, inputRequestName)
+	assert.Equal(false, inputWantReply)
+	wantJob := parseJob(t, wantJobJSON)
+	gotJob := parseJob(t, string(inputPayload))
+	assert.Equal(wantJob, gotJob)
```
matematik7, line 246: You could use `assert.JSONEq`, might instead of parsing the job.

matematik7 (dismissed):

client/exec.go
m-terel: done

client/exec_test.go
m-terel: done

matematik7 (approved):

PR #88, opened 2020-10-26
DEV-1751: Remote command execution documentation

m-terel:


rportd.example.conf
```diff
@@ -173,14 +173,14 @@
 [api]
   ## Defines the IP address and port the API server listens on
   ## specify non-empty {address} to enable API support
-  address = "127.0.0.1:9999"
+  address = "127.0.0.1:3000"
```
matematik7, line 176: Are these changes changes to rportd example config intentional?

docs/command-execution.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Deny any commands matching one of the following regular expressions
+## The filter is applied to the command sent.
+## With the below default filter only single commands are allowed.
+## Defaults: ['(\||<|>|;|\\|\n)']
+#deny = ['(\||<|>|;|\\\n)']
+
+## Order of which filter is applied first.
+## Defaults: ['deny','allow']
+#order = ['deny','allow']
```
matematik7, line 72: I don't really understand what the order would do, it would be nice to have some examples.

rportd.example.conf
m-terel: Yes, so examples from the docs would work with the default configuration. These default settings were incorrect because in all examples `127.0.0.1:3000` and `admin:foobaz` are used.

docs/command-execution.md
m-terel: This PR just adds the doc that was created by Thorsten. Next PR adds the output limitation. And one after that will implement this option. Let's discuss it in that PR, because I haven't clarified all requirements for myself yet. Currently, this section in this PR is marked as NOT IMPLEMENTED.

matematik7 (approved):

PR #89, opened 2020-10-26
DEV-1751: Limit size of command execution output

m-terel:


client/exec.go
```diff
@@ -151,6 +153,15 @@ func (c *Client) HandleRunCmdRequest(ctx context.Context, reqPayload []byte) (*c
 	return res, nil
 }
 
+func getJobResult(stdOut, stdErr bytes.Buffer, limit int) *models.JobResult {
+	stdOutBytes := stdOut.Next(limit)                    // cut the possible max bytes from stdout
+	stdErrBytes := stdErr.Next(limit - len(stdOutBytes)) // cut the rest of the limit from stderr
```
matematik7, line 158: I think this might not be the best behavior. Imagine a command filling stdout to the limit and then failing, you would get nothing on stderr and won't see the error.
I would just send up to limit of stdout and the same amount of stderr.

client/exec.go
```diff
@@ -151,6 +153,15 @@ func (c *Client) HandleRunCmdRequest(ctx context.Context, reqPayload []byte) (*c
 	return res, nil
 }
 
+func getJobResult(stdOut, stdErr bytes.Buffer, limit int) *models.JobResult {
+	stdOutBytes := stdOut.Next(limit)                    // cut the possible max bytes from stdout
```
matematik7, line 157: It would be better to limit the amount before the buffer, otherwise a command with huge output might OOM rport client.

client/exec_test.go
```diff
@@ -300,3 +312,124 @@ func TestHandleRunCmdRequestHasRunningCmd(t *testing.T) {
...
+		{
+			name:       "nonempty output zero limit",
+			stdOutStr:  testStdOut,
+			stdErrStr:  testStdErr,
+			limit:      0,
+			wantStdOut: "",
+			wantStdErr: "",
+		},
+		{
+			name:       "limit exceeds both stdout and stderr",
```
matematik7, line 373: I was a bit confused about the `exceeds` wording here, I think `is larger than` would be better.

client/exec.go
m-terel: Done.

client/exec.go
m-terel: Agree, this behavior was confusing to me too.

client/exec_test.go
m-terel: fixed

client/exec.go
```diff
@@ -179,3 +174,28 @@ var getShell = func(inputShell, os string) (string, error) {
...
+
+	// write to buffer only a part if exceeds the capacity
+	if len(p) > freeCapacity {
+		return b.buf.Write(p[:freeCapacity])
+	}
+
+	return b.buf.Write(p)
+}
+
+func (b *CapacityBuffer) String() string {
```
matematik7, line 199: Why not just embed the `bytes.Buffer` so `CapacityBuffer` would have all of it's methods and only override `Write`?

client/exec_test.go
```diff
@@ -237,28 +228,92 @@ func TestHandleRunCmdRequestNormalCase(t *testing.T) {
...
+			"stderr": "error1error2"
+		}
 	}
-}
+	`
+	ss := `output1output2output3`
+	stdOutSize := len(`output1output2output3`)
+	stdErrSize := len(`error1error2`)
+	fmt.Println(stdOutSize)
+	fmt.Println(stdErrSize)
```
matematik7: This seems like some leftover debugging code...

m-terel: Yor reviewed an outdated file.

client/exec.go
```diff
@@ -179,3 +174,28 @@ var getShell = func(inputShell, os string) (string, error) {
...
+	buf      bytes.Buffer
+	capacity int
+}
+
+func (b *CapacityBuffer) Write(p []byte) (n int, err error) {
+	freeCapacity := b.capacity - b.buf.Len()
+
+	// do not write to buffer if no space left
+	if freeCapacity <= 0 {
+		return 0, nil
```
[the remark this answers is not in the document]
m-terel: Done

client/exec.go
m-terel: done

matematik7 (approved):

PR #93, opened 2020-10-28
DEV-1645: Refactor users from file

matematik7:
- Remove unnecessary user repository
- Make sure reloading only happens when auth file is set for api
- Make reloading thread safe


m-terel (commented):
My suggestion is to refactor it using the same approach as before with having `UserProvider` interface and`CachedProvider` and other provider implementations. Would be great to have the same approach all over the project.

m-terel (approved):

matematik7:
@m-terel This one is a bit different since it uses the same cache for static and file auth and also does the reloading. I don't want to change it too much.

PR #95, opened 2020-10-29
DEV-1645: Make api auth required

matematik7:
Asked Thorsten and api auth should always be enabled.


server/config.go
```diff
@@ -182,6 +186,18 @@ func (c *Config) parseAndValidateAuth() error {
 	return nil
 }
 
+func (c *Config) parseAndValidateAPIAuth() error {
+	if c.API.AuthFile == "" && c.API.Auth == "" {
```
m-terel, line 190: Actually, API is not required to be enabled. This will fail when it's intentionally turned off. You should additionally check whether `c.API.Address` is set or not.

server/config.go
```diff
@@ -142,18 +142,22 @@ func (c *Config) ParseAndValidate() error {
 		return fmt.Errorf("expected 'Keep Lost Clients' can be in range [%v, %v], actual: %v", MinKeepLostClients, MaxKeepLostClients, c.Server.KeepLostClients)
 	}
 
-	if err := c.parseAndValidateAuth(); err != nil {
+	if err := c.parseAndValidateClientAuth(); err != nil {
 		return err
 	}
 
+	if err := c.parseAndValidateAPIAuth(); err != nil {
```
m-terel, line 149: And that check would be good to place here. Even better would be to refactor the code a little bit to move everything related to the API section to a separate func.

server/config.go
matematik7: Done.

server/config.go
matematik7: @m-terel Done.

m-terel (approved):

PR #97, opened 2020-11-05
DEV-1823: Migration to use sqlite: jobs

m-terel:


server/api/jobs/sqlite.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	_, err = p.db.NamedExec(`INSERT OR REPLACE INTO jobs (jid, status, started_at, finished_at, created_by, sid, command, shell, pid, timeout_sec, result)
+														VALUES (:jid, :status, :started_at, :finished_at, :created_by, :sid, :command, :shell, :pid, :timeout_sec, :result)`,
+		jobSqlite)
+	return err
+}
+
+type jobSqlite struct {
+	jobSummarySqlite
+	StartedAt  string         `db:"started_at"`
```
matematik7, line 81: I find it a bit weird that you're storging time objects as string.

matematik7 (dismissed):

server/api/jobs/sqlite.go
m-terel: @matematik7 afaik, sqlite doesn't have datetime type, that's why a string was used. https://www.sqlite.org/datatype3.html https://www.sqlite.org/lang_datefunc.html

Or do we still can use time object, use sqlite `TEXT` type (or create our own type) and scan would work? Ok, I'll take a look if time object can be used.

PR #162, opened 2021-02-25
DEV-1988: Do not keep obsolete clients after restart

m-terel:


m-terel:
Actually, obsolete clients are needed for re-establishing tunnels after server restart when keep-lost-clients is set to 0 or some very small value. But they would be shown on UI. To change it logic can be tuned to keep obsolete clients, but do not return them by APIs. It can be done in a separate PR.

PR #213, opened 2021-05-10
DEV-2124: 2F auth

m-terel:


server/api_listener.go
```diff
@@ -204,13 +221,16 @@ func (al *APIListener) handleAPIRequest(w http.ResponseWriter, r *http.Request)
 var ErrTooManyRequests = errors.New("too many requests, please try later")
 
 func (al *APIListener) lookupUser(r *http.Request) (authorized bool, username string, err error) {
```
breathbath, line 223: I know it was previously but it's probably not a big effort to change it, the thing is that method name doesn't correspond to what happens inside, lookup is a process of finding something rather than validating credentials or check users against blacklist. Maybe a name like `validateCredentials` or `validateUser` is more suitable here.

share/security/token.go
```diff
@@ -0,0 +1,18 @@
+package security
```
breathbath, line 1: I found this https://github.com/cloudradar-monitoring/rport/blob/master/share/random/random.go maybe this can be reused or this code can be placed there?

server/2fa_auth.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to generate 2fa token: %wv", err)
+	}
+
+	if err := srv.MsgSrv.Send(token, user.TwoFASendTo); err != nil {
+		return "", err
+	}
+
+	srv.mu.Lock()
+	srv.tokensByUser[username] = &expirableToken{
```
breathbath, line 75: I see it's a in memory database which holds all tokens so it will grow over time, maybe it's a wise idea to implement some basic cleanup procedure to remove expired tokens?

server/api/message/pushover.go
```diff
@@ -0,0 +1,50 @@
...
+		p: pushover.New(apiToken),
+	}
+}
+
+const pushoverAPISuccessStatus = 1
+
+func (s *PushoverService) Send(msg, receiver string) error {
+	pMsg := pushover.NewMessageWithTitle(msg, "Rport 2FA token")
+	pReceiver := pushover.NewRecipient(receiver)
+	resp, err := s.p.SendMessage(pMsg, pReceiver)
```
breathbath, line 28: no context option here? maybe it's not the best pushover library? passing timeouts to client calls would make our app more robust and less dependant from hanging server side?

server/api/message/pushover.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+
+	if resp.Status == pushoverAPISuccessStatus {
+		return nil
+	}
+
+	return errors.APIError{
+		Message: fmt.Sprintf("failed to send 2fa token, request: %s, status: %v, receipt: %s, errors: %v", resp.ID, resp.Status, resp.Receipt, resp.Errors),
+		Code:    http.StatusBadRequest,
```
breathbath, line 48: if pushover fails because of internal errors it should be status 500 here. as mentioned above we should probably have a more precise way to classify pushover response codes e.g. mirror http status codes from pushover

server/config.go
```diff
@@ -204,6 +231,26 @@ func (c *Config) parseAndValidateAPI() error {
...
+	if c.API.TwoFATokenDelivery == "" {
+		return nil
+	}
+
+	if c.API.Auth != "" {
+		return errors.New("2FA is not available if you use a single static user-password pair")
+	}
+
+	// TODO: to do better handling, maybe with using enums
+	switch c.API.TwoFATokenDelivery {
```
breathbath, line 244: I've seen same/similar logic in `server/api_listener.go` maybe it's good to have it once?

server/api/message/pushover.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/gregdel/pushover"
+
+	"github.com/cloudradar-monitoring/rport/server/api/errors"
+)
+
+type PushoverService struct {
```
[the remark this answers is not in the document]
m-terel: done

server/2fa_auth.go
```diff
@@ -0,0 +1,112 @@
...
+	return fmt.Sprintf("A token has been sent to %s", user.TwoFASendTo), nil
+}
+
+// TODO: add tests
+func (srv *TwoFAService) ValidateToken(username, token string) error {
+	srv.mu.RLock()
+	t := srv.tokensByUser[username]
+	defer srv.mu.RUnlock()
+
+	if t == nil {
```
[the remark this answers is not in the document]
breathbath: ok

server/api.go
```diff
@@ -396,6 +496,48 @@ func (al *APIListener) handleDeleteLogin(w http.ResponseWriter, req *http.Reques
...
+		return
+	}
+
+	successMsg, err := al.twoFASrv.SendToken(username)
+	if err != nil {
+		al.jsonError(w, err)
+		return
+	}
+
+	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(successMsg))
```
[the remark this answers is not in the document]
breathbath: yep - a good idea

PR #220, opened 2021-05-12
Revert "DEV-2124: Skip smtp config validation"

m-terel:
Reverts cloudradar-monitoring/rport#219


m-terel:
node1 wasn't working because of the smtp ports below 1024. Now it uses 2525 and it works ok.

PR #251, opened 2021-06-29
Additional os fields

breathbath:


server/api.go
```diff
@@ -565,21 +567,34 @@ func (al *APIListener) handleGetStatus(w http.ResponseWriter, req *http.Request)
...
+	var err error
 	sortFunc, desc, err := getCorrespondingSortFunc(req.URL.Query().Get(queryParamSort))
 	if err != nil {
 		al.jsonErrorResponse(w, http.StatusBadRequest, err)
 		return
 	}
 
-	clients, err := al.clientService.GetAll()
+	listOptions := query.ConvertGetParamsToFilterOptions(req)
+	err = query.ValidateListOptions(listOptions, clientsSupportedFields)
```
m-terel, line 578: Currently, if you use `?sort=name` it would fail with `unsupported sort field 'name'`. You are validating only filter options here. It might be good to have a func like `ValidateFilterOptions` to validate only filter options. Or extend `clientsSupportedFields` with fields that are available for sorting.

server/api.go
```diff
@@ -565,21 +567,34 @@ func (al *APIListener) handleGetStatus(w http.ResponseWriter, req *http.Request)
...
 	if err != nil {
-		al.jsonErrorResponse(w, http.StatusInternalServerError, err)
+		al.jsonError(w, err)
+		return
+	}
+
+	var cls []*clients.Client
+	if len(listOptions.Filters) > 0 {
+		cls, err = al.clientService.GetFiltered(listOptions)
```
m-terel, line 586: `GetFiltered` uses only filter options. It doesn't use sort options. If it's not used then please pass through only filter options.

server/api.go
```diff
@@ -661,44 +676,64 @@ func (al *APIListener) handleDeleteUser(w http.ResponseWriter, req *http.Request
...
-	IPv4            []string                `json:"ipv4"`
-	IPv6            []string                `json:"ipv6"`
-	Tags            []string                `json:"tags"`
-	Version         string                  `json:"version"`
-	Address         string                  `json:"address"`
-	Tunnels         []*clients.Tunnel       `json:"tunnels"`
-	DisconnectedAt  *time.Time              `json:"disconnected_at"`
-	ConnectionState clients.ConnectionState `json:"connection_state"`
-	ClientAuthID    string                  `json:"client_auth_id"`
+	NumCPUs                int                     `json:"num_cpus"`
```
m-terel, line 679: nit: API response has fields in the same order as they are declared here. It might be more user-friendly to have returned client payload with fields `id`, `name` at the top of the json response. Please declare fields `NumCPUs`, `MemoryTotal` and `DisconnectedAt` not as the top fields.

server/clients/client_builder.go
```diff
@@ -101,10 +112,13 @@ func (b ClientBuilder) Build() *Client {
...
-		ClientAuthID:   b.clientAuthID,
-
-		Connection: b.conn,
+		DisconnectedAt:        b.disconnectedAt,
+		ClientAuthID:          b.clientAuthID,
+		Connection:            b.conn,
+		Context:               nil,
+		Logger:                nil,
+		tunnelIDAutoIncrement: 0,
+		lock:                  sync.Mutex{},
```
m-terel, line 121: Please why do you explicitly set these private fields here? My suggestion to remove them.

server/clients/cr.go
```diff
@@ -189,3 +209,87 @@ func (s *ClientRepository) getNonObsolete() ([]*Client, error) {
...
+		}
+		if !matches {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+func (s *ClientRepository) clientMatchesFilter(cl *Client, filter query.FilterOption) (bool, error) {
```
m-terel, line 244: Do we have some tests that cover this func?

m-terel (commented):
@thorstenkramm @breathbath should new fileds be supported by https://petstore.swagger.io/?url=https://raw.githubusercontent.com/cloudradar-monitoring/rport/master/api-doc.yml#/Client%20Groups? It can be added in a separate PR.

client/client.go
```diff
@@ -391,6 +391,19 @@ func (c *Client) connectionRequest(ctx context.Context) *chshare.ConnectionReque
 		connReq.OS = "unknown"
 		c.Logger.Errorf("Could not get os name: %v", err)
 	}
+
+	connReq.OSFullName = c.getOSFullName(info)
```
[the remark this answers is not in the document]
breathbath: sorry but I am doing the tasks according the specifications, I don't know the reasons behind those ideas so I implement code like it's suggested

client/client.go
```diff
@@ -391,6 +391,19 @@ func (c *Client) connectionRequest(ctx context.Context) *chshare.ConnectionReque
...
+	if info != nil {
+		connReq.OSVersion = info.PlatformVersion
+	}
+
+	connReq.OSVirtualizationSystem, connReq.OSVirtualizationRole, err = c.systemInfo.VirtualizationInfo(ctx, info)
+	if err != nil {
+		connReq.OSFullName = "unknown"
```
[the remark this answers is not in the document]
breathbath: true - my bad

client/client.go
```diff
@@ -412,3 +452,15 @@ func (c *Client) getOS(ctx context.Context, info *host.InfoStat) (string, error)
 	}
 	return c.systemInfo.Uname(ctx)
 }
+
+func (c *Client) getOSFullName(infoStat *host.InfoStat) string {
+	if infoStat == nil {
+		return ""
```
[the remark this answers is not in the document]
breathbath: ok

client/system_info.go
```diff
@@ -48,3 +62,47 @@ func (s *realSystemInfo) InterfaceAddrs() ([]net.Addr, error) {
...
+		cpuInfo.NumCores = cpuCount
+	} else {
+		errs = append(errs, err2.Error())
+	}
+
+	return cpuInfo, errors.New(strings.Join(errs, ", "))
+}
+
+func (s *realSystemInfo) MemoryStats() (*mem.VirtualMemoryStat, error) {
+	return mem.VirtualMemory()
```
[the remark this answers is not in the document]
breathbath: ok

server/api.go
breathbath: why to be so strict? does it hurt somehow? I implemented it with the aim to add sorting in future

server/api.go
breathbath: it's for future

server/api.go
breathbath: no no - fields in structs should be sorted by the amount of memory they take (I think it was one of the linters which suggested me that in one of other projects) - this allows to have less fragmented memory consumption for the structs, as for order - it should be never guaranteed by the api

server/clients/client_builder.go
breathbath: does it hurt? I sometimes use autogeneration for structs and it is the result

server/clients/cr.go
breathbath: it's work in progress
I plan to add tests and adjust documentation

server/api.go
```diff
@@ -661,44 +676,64 @@ func (al *APIListener) handleDeleteUser(w http.ResponseWriter, req *http.Request
...
+	ID                     string                  `json:"id"`
+	Name                   string                  `json:"name"`
+	OS                     string                  `json:"os"`
+	OSArch                 string                  `json:"os_arch"`
+	OSFamily               string                  `json:"os_family"`
+	OSKernel               string                  `json:"os_kernel"`
+	Hostname               string                  `json:"hostname"`
+	OSFullName             string                  `json:"os_full_name"`
+	OSVersion              string                  `json:"os_version"`
+	OSVirtualizationSystem string                  `json:"os_virtualization_system"`
```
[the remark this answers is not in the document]
breathbath: it's on todolist

server/api.go
breathbath: here is some useful information:
https://medium.com/@sebassegros/golang-dealing-with-maligned-structs-9b77bacf4b97

m-terel: Thanks for the link. To be honest I've never had such lint errors. It could be really helpful if you are working with large data. But isn't it a premature optimization in our case? We don't have a system that manages millions of clients. If we had even few hundred clients and we save few bytes on each client then it wouldn't make much difference. Moreover, I've tried to structure fields in an order that I've proposed and `maligned` tool tells it would use the same number of bytes.

server/clients/client_builder.go
m-terel: These are private fields, if you change implementation it requires updating this code when it's not so necessary. If it was auto-generated then it would be clear why it's done so. But adding these fields manually is a little bit confusing.

client/client.go
```diff
@@ -401,6 +414,33 @@ func (c *Client) connectionRequest(ctx context.Context) *chshare.ConnectionReque
...
+
+	if err != nil {
+		c.Logger.Errorf("Could not get cpu information: %v", err)
+	}
+	if len(cpuInfo.CPUs) > 0 {
+		connReq.CPUFamily = cpuInfo.CPUs[0].Family
+		connReq.CPUModel = cpuInfo.CPUs[0].Model
+		connReq.CPUModelName = cpuInfo.CPUs[0].ModelName
+	}
+	if cpuInfo.NumCores > 0 {
```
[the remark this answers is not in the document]
breathbath: why - if it's -1?

server/api.go
breathbath: ok no issue for me - I can refactor, like you suggest
I tried that tool too but it reports too many issues so it might be hidden behind the reported ones

server/clients/client_builder.go
breathbath: already refactored

server/api.go
breathbath: I've added the check if sorting fields were provided it will report an error, I guess this is what you suggested? I don't want to argue to much about small things

server/api.go
breathbath: ok refactored

server/clients/cr.go
breathbath: done adding a test

server/api.go
breathbath: refactored

client/client.go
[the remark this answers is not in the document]
breathbath: done refactoring

server/api.go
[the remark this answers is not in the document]
breathbath: ok done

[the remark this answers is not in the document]
breathbath: ok done

server/api.go
m-terel: Nope, it's not what I meant. You had breaking changes. `handleGetClients` has the support of sorting https://github.com/cloudradar-monitoring/rport/blob/c2a2dc99761f04345bd16288bb30279c4180f6b3/server/api.go#L571

If a request contained `?sort=name` then your `ValidateListOptions` would return an error `unsupported sort field 'name'` because `clientsSupportedFields` doesn't contain `name` field. That's why my suggestion was to do one of:
- validate only filter options;
- add fields that are available for sorting to `clientsSupportedFields`.

You've extracted filter validation, so sorting works now as before.

m-terel (approved):

PR #252, opened 2021-07-01
DEV-2190: Limit client access

m-terel:


server/api.go
```diff
@@ -571,15 +607,33 @@ func (al *APIListener) handleGetClients(w http.ResponseWriter, req *http.Request
...
 	}
 
-	sortFunc(clients, desc)
+	curUser, err := al.getUserModelForAuth(req.Context())
+	if err != nil {
+		al.jsonError(w, err)
+		return
+	}
+
+	var filteredClients []*clients.Client
```
breathbath, line 622: maybe it's good to have all client filtering logic in client service rather than put it on top of existing clients list, I could imagine a use case where our clients are served from a database or even a cache and there you have to provide some query beforehand rather than stack filters on top of each other which might lead to unpredictable results

server/api.go
```diff
@@ -677,6 +731,7 @@ type ClientPayload struct {
 	DisconnectedAt  *time.Time              `json:"disconnected_at"`
 	ConnectionState clients.ConnectionState `json:"connection_state"`
 	ClientAuthID    string                  `json:"client_auth_id"`
+	ACL             []string                `json:"acl"`
```
breathbath, line 734: ACL is a service rather than data, maybe it's good to wrap acl logic in a service which holds some data so we can encapsulate acl complexity there?

server/api.go
```diff
@@ -745,6 +801,45 @@ func (al *APIListener) handleDeleteClient(w http.ResponseWriter, req *http.Reque
...
+	if cid == "" {
+		al.jsonErrorResponseWithTitle(w, http.StatusBadRequest, fmt.Sprintf("Missing %q route param.", routeParamClientID))
+		return
+	}
+
+	reqBody := &clientACLRequest{}
+	dec := json.NewDecoder(req.Body)
+	dec.DisallowUnknownFields()
+	err := dec.Decode(&reqBody)
+	if err == io.EOF { // is handled separately to return an informative error message
```
breathbath, line 820: maybe it's a good idea to make this logic reusable as I am pretty sure we have to handle this case in many other places

server/api.go
```diff
@@ -1594,11 +1689,29 @@ func (al *APIListener) handlePostMultiClientCommand(w http.ResponseWriter, req *
 		abortOnErr = *reqBody.AbortOnError
 	}
 
+	curUser, err := al.getUserModelForAuth(req.Context())
```
breathbath, line 1692: I would wrap access control to a separate service which can be unittested

server/api.go
```diff
@@ -1863,11 +1976,29 @@ func (al *APIListener) handleCommandsExecutionWS(ctx context.Context, uiConnTS *
...
 	}
 
+	curUser, err := al.getUserModelForAuth(ctx)
+	if err != nil {
+		uiConnTS.WriteError("Could not get current user.", err)
+		return
+	}
+
+	var clientsWithNoAccess []string
+	for _, curClient := range orderedClients {
```
breathbath, line 1986: looks like it's repeating on line 1699

server/api/users/service.go
```diff
@@ -53,6 +56,37 @@ func (as *APIService) GetAll() ([]*User, error) {
...
+	return nil, errors.New("not implemented")
+}
+
+func (as *APIService) ExistGroups(groups []string) error {
+	existingGroups, err := as.GetAllGroups()
+	if err != nil {
+		return err
+	}
+
+	groupSet := collections.ConvertToSet(existingGroups)
```
breathbath, line 72: just out of curiosity - why not do this filtering with a database query?

server/clients/client.go
```diff
@@ -222,6 +225,27 @@ func (c *Client) ConnectionState() ConnectionState {
...
+	}
+
+	return false
+}
+
+func (c *Client) DisplayName() string {
+	if c.Name != "" {
+		return c.Name
+	}
+	if len(c.IPv4) > 0 {
```
breathbath, line 243: and IPv6 are somehow not good enough?

server/clients/client.go
```diff
@@ -222,6 +225,27 @@ func (c *Client) ConnectionState() ConnectionState {
 	return Disconnected
 }
 
+func (c *Client) VerifyAccess(userGroups []string) bool {
```
breathbath, line 228: it's a conventional that if some method returns boolean it should be prefixed as "is" or "has", in this case maybe a good wording would be maybe "HasAccess"?

share/collections/set.go
```diff
@@ -0,0 +1,15 @@
+package collections
+
+type StringSet map[string]bool
```
breathbath, line 3: I think it's a hashmap rather than a stringset. Since go doesn't support sets, the wording is a bit confusing. I would call this type StringBoolMap or maybe Hashmap

share/collections/set.go
```diff
@@ -0,0 +1,15 @@
...
+
+func (s StringSet) Has(v string) bool {
+	return s[v]
```
breathbath, line 6: if I have here an element with a false value - has should return true as "set" contains the value. But maybe it's a philosophical interpretation ;-)

server/api.go
m-terel: I've moved client access check to client srv. IMO `getUserModelForAuth` can be in user service as smth like `GetUserFromContext(ctx)`.

server/api.go
breathbath: sure

server/api.go
breathbath: usually acl is quite complex maybe it could become more complex with time so it would be easier to find it, also it's will be possible to unittest it

server/api.go
breathbath: ok thanks

server/api.go
breathbath: ok

server/api.go
breathbath: ok

server/api/users/service.go
breathbath: but you can get unknown groups from database as well or do I miss something?

server/clients/client.go
breathbath: ok

share/collections/set.go
breathbath: yes but stringset cannot be backed by a hashmap where key is string ;-) it's a kind of slice with number indices and unique values, thanks for renaming ;-)

share/collections/set.go
breathbath: ok :-)

server/api/users/service.go
m-terel: I've remembered why I did it in this way. Because if the data source is a json file then this solution is common for both sources DB and file. I wanted to keep it simple. I know `GetGroups` return `"not implemented"` for file source now, but it will be available in the next PR where I merge our two existing user services.

server/api.go
m-terel: `CheckClientsAccess` is unit-tested. If it becomes more complex then we could easily extract it to a separate srv. I won't bother so much about it now :)

PR #253, opened 2021-07-05
Fixing powershell script execution

breathbath:


client/exec.go
```diff
@@ -126,6 +182,17 @@ func (c *Client) HandleRunCmdRequest(ctx context.Context, reqPayload []byte) (*c
 
 	// observe the cmd execution in background
 	go func() {
+		if job.IsScript {
```
m-terel, line 185: This goroutine is just an observing background task that would stop the observation after a given timeout. But the script can still be executing. That's why I'm not sure if we can delete the script file at this point.

breathbath: so what is your suggestion?

breathbath: the result is undefined, the deletion of the script will not change this status - it's still not clear if it fails or succeeds

m-terel: I don't know what happens exactly in such a scenario. If script execution is still in progress and we are trying to remove the file my assumption is that it can cause:
1. file deletion would fail and the script execution would continue;
2. file deletion would cause script execution failure.

We don't want any of that. I'm not sure that it would delete the file successfully and the script execution would successfully continue. It should be verified. If so then I'm ok with this solution. If not then we should come up with smth better.

I'm not familiar with powershell, but if I understood correctly, you have this fix because the command for script execution currently is built smth like: 
```powershell -Noninteractive -Command -executionpolicy bypass -file %s; powershell Remove-Item %s```
Can we try to build a command with `-file` only and without `-Command`? smth like
```powershell -Noninteractive -executionpolicy bypass -file %s; powershell Remove-Item %s```

Would it work?

breathbath: hey I've been trying to build a working command combined with deletion for the whole day - but nothing worked. We decided to use go to delete script and it worked. I think this is the cleanest solution. As for timeout situation assuming we don't delete script with go what would happen:
- script execution would continue and finish
- script execution would fail (because of hanging process or whatever)
both cases again happen no matter if you delete a file or not ;-)

m-terel (approved):

PR #257, opened 2021-07-08
DEV-2218: Add apt package manager for updates status

domenipavec:
This PR adds `updates` package to client with support for `apt` package manager. In future PRs I'll bring this data to server and support for other package managers.


client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+		descriptionRegex:       regexp.MustCompile(`\((.*?)\)`),
+		rebootRequiredFilename: "/var/run/reboot-required",
+		detectCmd:              []string{"apt-get", "help"},
+		updateCacheCmd:         []string{"sudo", "-n", "apt-get", "update", "-o", "Debug::NoLocking=true"},
+		getSummariesCmd:        []string{"apt-get", "-s", "-o", "Debug::NoLocking=true", "upgrade"},
+		getCountsCmd:           []string{"/usr/lib/update-notifier/apt-check"},
+	}
+}
+
+func (p *AptPackageManager) Detect() bool {
```
breathbath, line 33: sorry but it's not really clear to understand what Detect is doing if it returns bool, if a method returns a bool it should be have prefix either has or is to make it more readable

client/updates/runner.go
```diff
@@ -0,0 +1,32 @@
+package updates
+
+import (
+	"bytes"
+	"errors"
+	"os/exec"
+)
+
+type Runner interface {
```
breathbath, line 9: we already have a functionality to execute commands on the client, why we should build something on the top which will have same problems that we already resolved in our execution code?

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+	_, err := p.runner.Run(p.detectCmd...)
```
breathbath, line 34: I would say we should reuse exec code of client instead of inventing another "wheel"

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+	return err == nil
+}
+
+func (p *AptPackageManager) GetUpdatesStatus() *Status {
+	err := p.updatePackageCache()
+	if err != nil {
+		return &Status{
+			Error: err.Error(),
```
breathbath, line 42: it's a common pattern in go to return data separated from errors as there is no need for special error handling in custom models

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+
+	summaries, err := p.getSummaries()
+	if err != nil {
+		return &Status{
+			Error: err.Error(),
```
breathbath, line 49: same

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+}
+
+func (p *AptPackageManager) getCounts() (int, int, error) {
+	output, err := p.runner.Run(p.getCountsCmd...)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	parts := strings.SplitN(output, ";", 2)
```
m-terel, line 84: nit: I'm not sure if it can happen in practice, but maybe might be worth checking the length of `parts` to make sure we wouldn't get out of range.

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return &Status{
+			Error: err.Error(),
+		}
+	}
+
+	rebootPending, err := p.checkRebootRequired()
+	if err != nil {
+		return &Status{
+			Error: err.Error(),
```
breathbath, line 56: same

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+
+	// Get the number of available updates and the number of security updates
+	// from the output of the command, if is available. Ubuntu does some magic to calculate the updates.
```
breathbath, line 61: I am sorry why it's ubuntu? shouldn't this code also work on centos, alpine etc?

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+	getSummariesCmd        []string
+	getCountsCmd           []string
+}
+
+func NewAptPackageManager() *AptPackageManager {
+	return &AptPackageManager{
+		runner:                 &RunnerImpl{},
+		descriptionRegex:       regexp.MustCompile(`\((.*?)\)`),
+		rebootRequiredFilename: "/var/run/reboot-required",
+		detectCmd:              []string{"apt-get", "help"},
```
breathbath, line 26: isn't apt preferable over apt-get? anyway all commands should be in constants

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+		updateCacheCmd:         []string{"sudo", "-n", "apt-get", "update", "-o", "Debug::NoLocking=true"},
```
breathbath, line 27: sudo is not a part of command - this is an "option" on top of usual commands. Also if you run client as root there is no need for sudo -n

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+			Error: err.Error(),
+		}
+	}
+
+	// Get the number of available updates and the number of security updates
+	// from the output of the command, if is available. Ubuntu does some magic to calculate the updates.
+	// As this number is displayed on the system, we give it preference over our own sums.
+	// If not available, we count the summaries.
+	available, security, err := p.getCounts()
+	if err != nil {
```
breathbath, line 65: looks like error is ignored completely, what if the error doesn't indicate the missing counts but contains some internal go error like wrong inputs or code problems? I think we should not silently ignore any errors from our custom method but provide some logic to analyse the error type and detect the difference between errors coming from code from errors coming from external tools.

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return &Status{
+		UpdatesAvailable:         available,
+		SecurityUpdatesAvailable: security,
+		UpdateSummaries:          summaries,
+		RebootPending:            rebootPending,
+	}
+}
+
+func (p *AptPackageManager) getCounts() (int, int, error) {
```
breathbath, line 78: would be nice to indicate by named returned parameters what the first int should contain and what the second

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+type AptPackageManager struct {
```
breathbath, line 11: AptGetPackageManager as apt is another packagemanager which is not used here

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+	summaries, err := p.getSummaries()
+	if err != nil {
+		return &Status{
+			Error: err.Error(),
+		}
+	}
+
+	rebootPending, err := p.checkRebootRequired()
+	if err != nil {
+		return &Status{
```
m-terel, line 55: Assume `getSummaries ` was successful but `checkRebootRequired ` fails. Maybe instead of returning a status with one error, it makes sense to update the `status` on the fly and proceed further? I mean we just set all successful results there and append all errors. What do you think?

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, err
+	}
+
+	var result []Summary
+	for _, line := range strings.Split(output, "\n") {
+		if !strings.HasPrefix(line, "Inst") {
+			continue
+		}
+
+		title := strings.SplitN(line, " ", 3)[1]
```
m-terel, line 130: Might be good to check the length to avoid index out of range.

client/updates/apt.go
```diff
@@ -0,0 +1,151 @@
...
+		description := fmt.Sprintf(
+			"%s %s",
+			title,
+			p.descriptionRegex.FindStringSubmatch(line)[1],
```
m-terel, line 134: Same here.

client/updates/updates.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+type Summary struct {
+	Title            string
+	Description      string
+	RebootRequired   bool
+	IsSecurityUpdate bool
+}
+
+func New(logger *chshare.Logger, interval time.Duration) *Updates {
```
m-terel, line 43: It might be good to propagate context here and further. So we could use it and be able to stop the execution based on timeouts/cancelation/etc.

client/updates/updates.go
```diff
@@ -0,0 +1,110 @@
...
+
+		logger: logger,
+	}
+
+	packageManagers := []PackageManager{
+		NewAptPackageManager(),
+	}
+	for _, pm := range packageManagers {
+		if pm.Detect() {
+			u.pkgMgr = pm
```
breathbath, line 56: what if I have an ubuntu without a package manager? I think this is an error case but this code silently ignores such cases. As we have knowledge about which os our clients have we can try to use a  package manager specific to the client's os rather than try to execute all possible package managers against all possible OS versions

client/updates/updates.go
```diff
@@ -0,0 +1,110 @@
...
+				u.logger.Infof("Update status refreshed, %v updates available (%v security)",
+					status.UpdatesAvailable, status.SecurityUpdatesAvailable)
+			}
+		}
+	}
+}
+
+func (u *Updates) refreshStatus() {
+	var newStatus *Status
+	if u.pkgMgr == nil {
```
m-terel, line 86: What if we move this check to `New` func and do not start the loop if there is no manager? Wouldn't it be better, what do you think?

client/updates/updates.go
```diff
@@ -0,0 +1,110 @@
...
+	newStatus.Refreshed = time.Now()
+
+	u.mtx.Lock()
+	defer u.mtx.Unlock()
+
+	u.status = newStatus
+}
+
+func (u *Updates) GetStatus() *Status {
+	u.mtx.Lock()
```
m-terel, line 102: nit: we could use RWMutex and RLock here.

breathbath, line 102: maybe it's better to use RWMutex here since reads should not block writes which might give a better performance

breathbath (commented):
I am not sure about the context of this code but I didn't find any usages of it in the existing code - is it some prototype?

client/updates/updates.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	go u.refreshLoop()
+	return u
+}
+
+func (u *Updates) refreshLoop() {
+	u.refreshStatus()
+	for {
+		select {
```
m-terel, line 67: nit: We can do this block before select:
https://github.com/cloudradar-monitoring/rport/blob/bb625debcb6ab653a1eabe463f3042dc4413576d/client/updates/updates.go#L71-L79
then we wouldn't need to do refresh before `for` loop (line 65).

client/updates/runner.go
```diff
@@ -0,0 +1,32 @@
...
+	Run(...string) (string, error)
+}
+
+type RunnerImpl struct{}
+
+func (r *RunnerImpl) Run(args ...string) (string, error) {
+	stderr := &bytes.Buffer{}
+	stdout := &bytes.Buffer{}
+
+	cmd := exec.Command(args[0], args[1:]...)
```
m-terel, line 19: It might be good to propagate ctx here and use `CommandContext`.

m-terel (approved):
Overall looks good, some minor comments.

client/updates/apt.go
domenipavec: Good point, I renamed it to `IsAvailable` - it checks whether the package manager is present on the system.

client/updates/runner.go
domenipavec: Executing general purpose commands and running specific package manager commands are quite different use cases. I can easily imagine adding something to one that would have unwanted consequences for the other.
I don't want to couple the code for a purpose of saving duplication of a few lines of code. Also the only reason I'm using interface here is so I can swap it in tests.

client/updates/apt.go
domenipavec: Duplicating a few lines of code is hardly reinventing the wheel. Also I made the interface here very specific to the use case (e.g. returning stdout as string for purposes of parsing the output and stderr in error), which makes it very easy to write tests.

client/updates/apt.go
domenipavec: You're right, I was writing against the specification of having the json with possible error field as output to send to server, but it makes this internal code non standard go. Changed to return status and error.

client/updates/apt.go
domenipavec: Done

client/updates/apt.go
domenipavec: Done

client/updates/apt.go
domenipavec: As the comment specifies, if the command is available (that is mostly on ubuntu based systems, since the used command is from ubuntu), we use that for the counts, otherwise (on other systems) we manually count updates and security updates.

client/updates/apt.go
domenipavec: My understanding is that apt is ubuntu's wrapper over a lower level apt-get command. Since we're parsing the output I think the lower level apt-get is perferable. It's also going to be more widely available then apt that might not be present on some systems.

As for putting them in constants - they cannot be constants since they're slices. Also I wanted to namespace them to apt, that's why I put them in the struct.

client/updates/apt.go
domenipavec: Running client as root is not recommended by our documentation. Since this command cannot work without super user privileges (and others can) I would argue that sudo is part of this command. As part of the documentation for this functionality there will be instructions how to setup sudoers file to make this work.

client/updates/apt.go
domenipavec: You're right, added a check if it's a command error than we ignore it, but if it's our code's error than we log it. I don't want to fail, since we can still count manually, but you're right we should log the error.

client/updates/apt.go
domenipavec: Good idea :+1:

client/updates/apt.go
domenipavec: Package manager's name is APT (Advanced Package Tool). It can be interacted with either apt-get or apt commands. I think the naming is fine.

client/updates/updates.go
domenipavec: Running ubuntu without a package manager should be a valid option, in that case the detection will fail and we will report that there is no supported package manager.
I think it's simpler to just detect if each package manager is available rather than construct a complex mapping from all possible OS versions to their package manager (a mapping we would have to maintain and update in the future). There are probably over 10 distributions that use apt and new ones could be developed at any time.

client/updates/updates.go
domenipavec: While checking for updates every few hours is hardly a performance task, I guess sure since you both commented on it.

client/updates/apt.go
domenipavec: Good point, since the output is of external command we should definitely check.

client/updates/apt.go
domenipavec: Since `checkRebootRequired` only checks for presence of a file it's very unlikely to fail. I don't think it's worth complicating the implementation for this.

client/updates/apt.go
domenipavec: Done

client/updates/apt.go
domenipavec: Done

client/updates/updates.go
domenipavec: Good idea, it removes need for custom stop channel functionality.

client/updates/updates.go
domenipavec: Good idea :+1:

client/updates/updates.go
domenipavec: Done

client/updates/updates.go
domenipavec: Not sure how I missed this, excellent idea :+1:

client/updates/runner.go
domenipavec: Done

domenipavec:
> I am not sure about the context of this code but I didn't find any usages of it in the existing code - is it some prototype?

This is part of the task to bring updates information to the server. This PR is mostly focused on getting data for apt package manager, future PRs will bring the data to the server and implement for other package managers.

client/updates/apt.go
```diff
@@ -30,39 +37,36 @@ func NewAptPackageManager() *AptPackageManager {
...
 	}
 
 	// Get the number of available updates and the number of security updates
 	// from the output of the command, if is available. Ubuntu does some magic to calculate the updates.
 	// As this number is displayed on the system, we give it preference over our own sums.
 	// If not available, we count the summaries.
-	available, security, err := p.getCounts()
+	available, security, err := p.getCounts(ctx)
 	if err != nil {
+		if _, ok := err.(getCountsCmdError); !ok {
```
m-terel, line 65: Maybe it might be useful to log the `getCountsCmdError` as well, but with `info/debug` level.

m-terel (approved):

PR #263, opened 2021-07-12
Disable script option

breathbath:


docs/no06-command-execution.md
```diff
@@ -190,6 +190,11 @@ On the client using the `rport.conf` you can configure and limit the execution o
 ## Defaults: true
 #enabled = true
 
+[remote-scripts]
+## Enable or disable remote scripts.
+## Defaults: false
+#enabled = false
```
domenipavec, line 196: Won't this break the options below since they would be in wrong section?

client/exec.go
```diff
@@ -80,6 +80,10 @@ func (c *Client) HandleRunCmdRequest(ctx context.Context, reqPayload []byte) (*c
 		c.Debugf("Waiting for a previous command with PID %d to finish", *curPID)
 	}
 
+	if job.IsScript && !c.config.RemoteScripts.Enabled {
```
domenipavec, line 83: Is there a usecase where remote scripts are enabled but remote commands are disabled? Because the check at the top of this function would also disable scripts.
Also it would be nicer to have the checks for scripts and commands being enabled together.

domenipavec (approved):

client/exec.go
m-terel, line 83: Also, in the current impl if `remote-scripts` is disabled the server would still send the script file to the client and only after that it would fail the execution. So the script file wouldn't be deleted from the client. Ideally, we should fail it earlier or don't send the file to the client at all.

One of the solution could be to include `remote-scripts` config property

docs/no06-command-execution.md
breathbath: it's a completely wrong place - I've moved it to 
no14-scripts.md

client/exec.go
breathbath: it won't as there is a check in the file handler to reject script files if the script execution is disabled
see 
https://github.com/cloudradar-monitoring/rport/pull/263/files#diff-da1faa2bf9b664f7e90deb0b7067b287a1def717f0b61e54db9efc21a6e713f6R24

domenipavec (approved):

PR #266, opened 2021-07-20
DEV-2218: Endpoint to trigger updates status refresh on the client

domenipavec:


api-doc.yml
```diff
@@ -539,6 +539,34 @@ paths:
           description: "Invalid Operation"
           schema:
             $ref: "#/definitions/ErrorPayload"
+  /clients/{client_id}/updates_status:
```
m-terel, line 542: nit: to be consistent with other APIs maybe use `-` instead `_`

client/updates/updates.go
```diff
@@ -28,16 +28,18 @@ type Updates struct {
 	conn   ssh.Conn
 	status *models.UpdatesStatus
 
-	interval time.Duration
+	interval    time.Duration
+	refreshChan chan struct{}
```
m-terel, line 32: nit: Sorry, why did you use `struct{}`? Maybe just `bool`?

server/api.go
```diff
@@ -180,6 +180,7 @@ func (al *APIListener) initRouter() {
 	sub.HandleFunc("/clients/{client_id}/commands", al.wrapClientAccessMiddleware(al.handleGetCommands)).Methods(http.MethodGet)
 	sub.HandleFunc("/clients/{client_id}/commands/{job_id}", al.wrapClientAccessMiddleware(al.handleGetCommand)).Methods(http.MethodGet)
 	sub.HandleFunc("/clients/{client_id}/scripts", al.wrapClientAccessMiddleware(al.handleExecuteScript)).Methods(http.MethodPost)
+	sub.HandleFunc("/clients/{client_id}/updates_status", al.wrapClientAccessMiddleware(al.handleRefreshUpdatesStatus)).Methods(http.MethodPost)
```
m-terel, line 183: Just to clarify, is this API should be accessible only by admins?

server/api.go
```diff
@@ -2662,3 +2663,30 @@ func parseRequestBody(reqBody io.ReadCloser, dest interface{}) error {
 
 	return nil
 }
+
+func (al *APIListener) handleRefreshUpdatesStatus(w http.ResponseWriter, req *http.Request) {
+	vars := mux.Vars(req)
+	clientID, exists := vars[routeParamClientID]
+	if !exists || clientID == "" {
```
m-terel, line 2670: nit: I would say `exists` can be dropped here :)

m-terel (approved):

api-doc.yml
domenipavec: Sure, I didn't realize that was used.

client/updates/updates.go
domenipavec: I think it's better to use empty struct which has only 1 possible value, as that makes it clear that value doesn't matter. With `bool` I would assume that behavior is different whether it's true or false, which here it isn't.
Also, not that it matters here much, but `struct{}` has size 0, so it takes less space than `bool` :)

server/api.go
domenipavec: While, it's not really specified in the ticket, I would assume that anyone who can see the client can also refresh it's update status. Based on usage of the middleware I assume that's the same as for the other methods?

server/api.go
domenipavec: Sure, since I copied this from above, I cleaned up this pattern throughout the file.

server/api.go
m-terel: Oh, sorry, I confused it with admin access, all good

server/api.go
m-terel: I see, thanks :)

m-terel (dismissed):

m-terel (dismissed):

m-terel (approved):

PR #268, opened 2021-07-22
DEV-2218: Add yum package manager support

domenipavec:


client/updates/updates.go
```diff
@@ -61,6 +63,7 @@ func (u *Updates) Start(ctx context.Context) {
 		}
 		u.logger.Errorf("Update status not available: %v", u.status.Error)
 	} else {
+		u.logger.Infof("Using %v for updates", reflect.TypeOf(u.pkgMgr).Elem().Name())
```
m-terel, line 66: nit: it might be better to use `%T` instead of reflection

client/updates/yum.go
```diff
@@ -0,0 +1,118 @@
...
+		p.cmd = cmd
+		_, err := p.run(ctx, "help")
+		if err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (p *YumPackageManager) GetUpdatesStatus(ctx context.Context, logger *chshare.Logger) (*models.UpdatesStatus, error) {
```
m-terel, line 36: nit: `logger` is not used here, maybe just make it as a part of the manager, to avoid passing it around as func arg.

client/updates/yum.go
```diff
@@ -0,0 +1,118 @@
...
+	// exit code 100 means updates available, so continue
+	if err != nil && err.Error() != "exit status 100" {
+		return nil, err
+	}
+
+	result := []yumUpdate{}
+	for _, line := range strings.Split(output, "\n") {
+		parts := strings.Fields(line)
+		if len(parts) != 3 {
+			continue
```
m-terel, line 89: nit: It might be useful at least to log such lines.

client/updates/yum.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		result = append(result, update)
+	}
+
+	return result, nil
+}
+
+func (p *YumPackageManager) checkRebootRequired(ctx context.Context) bool {
+	output, _ := p.run(ctx, "needs-restarting", "-r")
```
m-terel, line 111: Would be good not to silence the error, return it and handle it higher.

client/updates/yum_test.go
```diff
@@ -0,0 +1,182 @@
...
+			ListUpdatesErr: errors.New("some error"),
+			ExpectedError:  errors.New("some error"),
+		},
+		{
+			Name:                   "List security updates error",
+			ListSecurityUpdatesErr: errors.New("some error"),
+			ExpectedError:          errors.New("some error"),
+		},
+		{
+			Name: "No security updates",
```
m-terel, line 99: Would be great to have a test with security updates as well.

m-terel (approved):

client/updates/updates.go
domenipavec: I've tried that, but I find this format a bit nicer, since `%T` also includes package name.

client/updates/yum.go
domenipavec: The way managers are initialized at the moment it's a bit hard to pass it there, so I'll leave it like this for now.

client/updates/yum_test.go
domenipavec: Not sure how I missed that, I clearly prepared `ListSecurityUpdatesOutput` for just this test :)

client/updates/yum.go
m-terel: Please add an explanatory comment on why the err is discarded.

PR #269, opened 2021-07-23
DEV-2218: Add zypper package manager support

domenipavec:


client/updates/zypper.go
```diff
@@ -0,0 +1,200 @@
...
+	needsRebootCmd []string
+	listPatchesCmd []string
+	patchInfoCmd   []string
+}
+
+func NewZypperPackageManager() *ZypperPackageManager {
+	return &ZypperPackageManager{
+		runner: &RunnerImpl{},
+
+		detectCmd:      []string{"zypper", "help"},
```
m-terel, line 27: Please tell why are these cmds as part of  `ZypperPackageManager`? Can we just use global vars?

client/updates/zypper.go
```diff
@@ -0,0 +1,200 @@
...
+		updateCacheCmd: []string{"sudo", "-n", "zypper", "refresh"},
+		listUpdatesCmd: []string{"zypper", "--terse", "--quiet", "list-updates"},
+		needsRebootCmd: []string{"zypper", "needs-rebooting"},
+		listPatchesCmd: []string{"zypper", "--terse", "--quiet", "list-patches", "-a"}, // TODO: remove -a
```
m-terel, line 31: Sorry, what is this todo about? Should `-a` be removed? :)

client/updates/zypper.go
```diff
@@ -0,0 +1,200 @@
...
+	listPatchesCmd []string
+	patchInfoCmd   []string
+}
+
+func NewZypperPackageManager() *ZypperPackageManager {
+	return &ZypperPackageManager{
+		runner: &RunnerImpl{},
+
+		detectCmd:      []string{"zypper", "help"},
+		updateCacheCmd: []string{"sudo", "-n", "zypper", "refresh"},
```
m-terel, line 28: nit: Can we do smth similar to https://github.com/cloudradar-monitoring/rport/blob/346f725000edfa1bc723d643d4ccf81426bc23bb/client/exec_nix.go#L13-L16

or if it makes sense just reuse one cmd executor/runner?

client/updates/zypper.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (p *ZypperPackageManager) IsAvailable(ctx context.Context) bool {
+	_, err := p.runner.Run(ctx, p.detectCmd...)
+	return err == nil
+}
+
+func (p *ZypperPackageManager) GetUpdatesStatus(ctx context.Context, logger *chshare.Logger) (*models.UpdatesStatus, error) {
+	err := p.updatePackageCache(ctx)
+	if err != nil {
+		return nil, err
```
m-terel, line 44: nit: maybe it would be good to add some context to each returned error.

client/updates/zypper.go
```diff
@@ -0,0 +1,200 @@
...
+	needsReboot bool
+}
+
+type zypperUpdate struct {
+	name    string
+	arch    string
+	version string
+}
+
+func (p *ZypperPackageManager) getSecurityAndRebootInfo(ctx context.Context) (map[string]zypperInfo, error) {
```
m-terel, line 127: nit: maybe it would be great to have some docs/explanatory comments to functions like this where it would have a short description of what the expected output is and how it's parsed. It would help a lot to quickly understand the logic.

client/updates/zypper.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		result = append(result, parts[0])
+	}
+
+	return result, nil
+}
+
+func (p *ZypperPackageManager) checkRebootRequired(ctx context.Context) bool {
+	output, _ := p.runner.Run(ctx, p.needsRebootCmd...)
```
m-terel, line 193: nit: please add an explanatory comment why the error is not handled.

m-terel (approved):

client/updates/zypper.go
domenipavec: Done

client/updates/zypper.go
domenipavec: Done

client/updates/zypper.go
domenipavec: Since some commands here always require `sudo` and some never need it, I think it's best to just include that as part of command.

client/updates/zypper.go
domenipavec: Thanks for catching this, I used `-a` for testing since it shows some demo patches.

client/updates/zypper.go
domenipavec: Since we have different package managers implementations I thought it would be nice to scope commands inside the structs. Since these are slices we can't use `const` anyway, then it doesn't make much difference if they're global or not.

m-terel (approved):

PR #272, opened 2021-07-27
DEV-2218: Add support for windows updates

domenipavec:


client/updates/updates.go
```diff
@@ -18,6 +18,7 @@ var packageManagers = []PackageManager{
 	NewZypperPackageManager(),
 	NewYumPackageManager(),
 	NewAptPackageManager(),
+	NewWindowsPackageManager(),
```
m-terel, line 21: Maybe it might be better to have `packageManagers` for nix and for windows. It would allow to eliminate unnecessary `WindowsPackageManager` nix impl.

m-terel (commented):
Would be good if we could cover it with some tests.

client/updates/updates.go
domenipavec: Done

domenipavec:
@m-terel I've thought quiet a bit about tests, and I'm not sure they're worth the effort here:
1. One option is to write a high level interface over `go-ole` which would only have high level methods like `listUpdates`. That would be relatively easy to write, but tests based on that would only cover minimal amount of code that would be needed around that.
2. The other option is to write interfaces for `go-ole` directly, which would be pretty complicated. And I don't think the effort would be worth it, because the tests assertions would be very tightly coupled to the implementation. I don't like such tests very much because they just copy the same thing that's already in the code.

breathbath (approved):

PR #275, opened 2021-08-04
DEV-2246: 2FA using script

domenipavec:


docs/no02-api-auth.md
```diff
@@ -51,15 +52,35 @@ By default, 2FA is disabled.
    two_fa_token_delivery = 'smtp'
    two_fa_token_ttl_seconds = 600
    ```
-   Use either `'smtp'` or `'pushover'`.
+   Use either `'smtp'`, `'pushover'` or provide a path to a binary or script executable.
```
breathbath, line 55: so basically 2fa script can be used to execute any scripts/commands? but since we already have an implemented way to execute scripts and commands we should not repeat the same in this interface otherwise you should also specify here parameters like sudo, shell, cwd, timeout...
So maybe it's good to just limit a custom method of sending 2fa code to only http api calls (via build in golang http client)?

server/api/message/script.go
```diff
@@ -0,0 +1,80 @@
...
+
+func NewScriptService(script string, validation ValidationType, regex *regexp.Regexp) *ScriptService {
+	return &ScriptService{
+		Script:     script,
+		Validation: validation,
+		Regex:      regex,
+	}
+}
+
+func (s *ScriptService) Send(ctx context.Context, data Data) error {
```
breathbath, line 38: so as mentioned above this repeats our execution logic for the commands and scripts which is not good (because there we already solved most of the issues related to e.g. windows executions but here we probably miss them + double effort if you need to adjust some execution bugs e.g. related to the non-utf8 output from cmd.exe etc)

server/api/message/script.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return nil
+}
+
+func (s *ScriptService) DataToEnv(data Data) []string {
+	return []string{
+		fmt.Sprintf("2FA_TOKEN=%v", data.Token),
+		fmt.Sprintf("2FA_SENDTO=%v", data.SendTo),
+		fmt.Sprintf("2FA_TOKEN_TTL=%v", data.TTL.Seconds()),
```
breathbath, line 60: for such kind of parameters you could use custom strings like 30s, 10m, 1h and parse them with duration parse logic of go

server/api/message/service.go
```diff
@@ -1,9 +1,20 @@
...
 
-import "context"
+import (
+	"context"
+	"time"
+)
+
+type Data struct {
+	Title   string
+	Message string
```
breathbath, line 10: so I have message.Data struct which is probably some data for message but then I have also message.Data.Message with stutters a bit, maybe it's good to use here something like text or body to make it more readable?

server/api/message/service.go
```diff
@@ -1,9 +1,20 @@
...
+	SendTo  string
+	Token   string
```
breathbath, line 12: I is single string field covers all possible ways to auth to an external API? what if login and a password should be used? maybe it's good to have also a set of dynamic fields for such cases e.g. map[string]interface{} so you can pass any parameters of any type through this interface?

breathbath (approved):

breathbath:
So generally I have no objections to the code itself but to the implementation idea. If this stays as it is please use our executor logic rather than implement it yourself. Thanks

docs/no02-api-auth.md
domenipavec: The ticket was specifically to make this work with scripts, which makes it more flexible since we don't have to implement support for all different APIs, authentication methods, etc. in our code. Also users might use something different that is not an HTTP call.

server/api/message/script.go
domenipavec: Our current logic is all implemented in client, this is for server and has very specific requirements (no sudo, working directory or arguments and our params in environment). We don't really capture output of the command, just stderr in case of error.

server/api/message/script.go
domenipavec: The code uses go duration, but here it's passed to the third-party executable that's probably not in go, so it wouldn't make sense to use go's duration format. So using seconds is much cleaner I think.

server/api/message/service.go
domenipavec: The token is not used to authenticate to external API, it's our 2FA token that is just passed to the script. And it is always a string.

server/api/message/service.go
domenipavec: The argument has been named message previously, that's why I don't want to change it. Since you would never actually write `message.Data.Message`, I don't see a problem.

domenipavec:
@breathbath The implementation idea has been specified in the ticket and it provides a lot of flexibility to the user to handle the 2fa token however they want.
As to the reusing the executor logic - I think go's exec is simple enough, that the code here is pretty simple. I don't think it's worth reusing the client's code here in server and coupling the implementation when we don't need any of that functionality.

PR #276, opened 2021-08-10
DEV-2229 Script shebang dir

breathbath:


client/exec_test.go
```diff
@@ -129,91 +129,112 @@ const scriptToRunJSON = `
...
-				assert.Equal(t, tc.wantShell, gotShell)
+				assert.Equal(t, tc.wantInterpreter, gotInterpreter)
 			}
 		})
 	}
 }
 
 func TestHandleRunCmdRequestPositiveCase(t *testing.T) {
 	now = nowMockF
-	assert := assert.New(t)
```
domenipavec, line 212: I don't think this is a common pattern to abbreviate it, the `assert` docs show using it as `assert`.
Also it only has 132 uses in whole github: https://github.com/search?q=%22asrt+%3A%3D+assert.new%28t%29%22&type=code :D

client/file.go
```diff
@@ -43,16 +48,16 @@ func (c *Client) HandleCreateFileRequest(ctx context.Context, reqPayload []byte)
...
-	baseDir := filepath.Dir(fileInput.Name)
-	if baseDir == "." || baseDir == string(os.PathSeparator) {
-		baseDir = os.TempDir()
-		fileInput.Name = filepath.Join(baseDir, filepath.Base(fileInput.Name))
-	} else if fileInput.CreateDir {
-		err = os.MkdirAll(baseDir, DefaultDirMode)
+	scriptFileName := filepath.Base(fileInput.Name)
+	scriptDirName := c.config.RemoteScripts.Dir
+
+	if fileInput.CreateDir {
```
domenipavec, line 54: I find it weird that the input can specify `CreateDir`, if the dir is always the one specified in the config.
Also wouldn't `ValidateScriptDir` above fail if directory doesn't exist, making this redundant?

client/file_test.go
```diff
@@ -26,9 +26,15 @@ func TestHandleCreateFileRequest(t *testing.T) {
 	inputFileBytes, err := json.Marshal(inputFile)
 	require.NoError(t, err)
 
+	const scriptDirToCheck = "TestHandleCreateFileRequestDir"
```
domenipavec, line 29: The name for this is a bit weird, we're not actually checking the dir?
Also I prefer to use go's temp file functionality to create tmp folder in tests instead of creating it in current working dir.

client/script_test.go
```diff
@@ -0,0 +1,108 @@
...
+			shouldCreateDir: true,
+		},
+		{
+			name:             "file as dir name",
+			dirToGive:        "some_file",
+			shouldCreateFile: true,
+			errToExpect:      "script_dir some_file is not a valid directory",
+		},
+		{
+			name:            "wrong_dir_mode",
```
domenipavec, line 57: Maybe split the tests the same as actual implementation and put this behind build flag? In case we run tests on windows these would fail I think.

client/script_nix.go
```diff
@@ -0,0 +1,57 @@
...
+			"script_dir %s must be read-writable only by %s[%s]. Change directory mode from 0%o to 0%o. Your setup is insecure",
+			scriptDir,
+			curUser.Username,
+			curUser.Uid,
+			dirMode,
+			DefaultDirMode,
+		)
+	}
+
+	scriptDirOwner, ok := scriptDirSysInfo.Sys().(*syscall.Stat_t)
```
domenipavec, line 40: Maybe rename this to `unixScriptDirStat`?

rport.example.conf
```diff
@@ -164,3 +164,14 @@ auth = "clientAuth1:1234"
   ## Enable or disable execution of remote scripts sent by server.
   ## Defaults: false
   #enabled = false
+  ##
+  ## Temporary directory to store scripts sent by the server before execution.
+  ## This directory must exist and it should be owned only by the rport user.
```
domenipavec, line 169: Considering we validate this, I would change should to must.

domenipavec (dismissed):
A few minor comments, but otherwise LGTM

client/config.go
```diff
@@ -212,6 +215,12 @@ func (c *Config) parseRemoteScripts() error {
 		return errors.New("remote scripts execution requires remote commands to be enabled")
 	}
 
+	err := ValidateScriptDir(c.RemoteScripts.Dir)
```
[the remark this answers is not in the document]
breathbath: it was explicitly required to validate scripts dir like this

client/exec_nix.go
```diff
@@ -9,15 +9,18 @@ import (
 
 func (e *CmdExecutorImpl) New(ctx context.Context, execCtx *CmdExecutorContext) *exec.Cmd {
 	var args []string
-	prefix := execCtx.Shell
 	if execCtx.IsSudo {
-		prefix = "sudo"
-		args = append(args, "-n", execCtx.Shell)
+		args = append(args, "sudo", "-n")
```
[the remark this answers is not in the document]
breathbath: because if I move this code part down the line I don't want to change the code

client/exec_test.go
breathbath: I am sorry I don't follow - this line is deleted

client/file.go
breathbath: because this api can be used in other cases like create files in directories which don't exist

client/file_test.go
breathbath: ok will use the temp dir

client/script_test.go
breathbath: I don't think our tests will pass windows without serious modifications but anyway I'll exclude this test from windows

client/script_nix.go
breathbath: sure

rport.example.conf
breathbath: must is anyway there but I'll remove should if you like

client/config.go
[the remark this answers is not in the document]
domenipavec: Ok. Another thing is it might be confusing for users that this is not part of the other logs, maybe it would be better to move this to use our logger?

client/exec_test.go
domenipavec: Sorry this was meant to be on the added side of the diff. You changed this line to `asrt := assert.New(t)`.
I don't think that's a common pattern to abbbreviate it like that.

rport.example.conf
domenipavec: I would change it to `This directory must exist and it must be owned only by the rport user`. Because both conditions must be true for our validation to pass.

client/file.go
domenipavec: I don't think this code can ever create directory, because above you're using `ValidateScriptDir` and returning if the dir doesn't exist.

client/exec_test.go
breathbath: but you cannot name variables same as packages - so therefore it's renamed

client/config.go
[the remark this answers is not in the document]
breathbath: what line are you referring to?

client/exec_test.go
breathbath: ok I'll use our common pattern for calling assertions

client/file.go
breathbath: ok fair - I'll remove it

rport.example.conf
breathbath: ok

client/exec_test.go
domenipavec: It's a common pattern to name variables same as package for this particular use case, but I agree it's a bit weird.

client/config.go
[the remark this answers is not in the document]
domenipavec: I'm refering to the log for this validation, you're using go's `log`, while we use our custom logger for all the other logs (and that might go to file, while this would still go to stderr I think).

domenipavec (approved):

PR #277, opened 2021-08-13
DEV-2250: Connect to fallback server if main is unavailable

domenipavec:
Connect to fallback servers if main is unavailable.
Try to switchback to the main server every interval.

https://tracker.cloudradar.info/issue/DEV-2250


client/client.go
```diff
@@ -135,119 +131,206 @@ func (c *Client) connectionLoop(ctx context.Context) {
...
+		ReadBufferSize:   1024,
+		WriteBufferSize:  1024,
+		HandshakeTimeout: 45 * time.Second,
+		Subprotocols:     []string{chshare.ProtocolVersion},
+	}
+	//optionally proxy
+	if c.config.Client.proxyURL != nil {
+		if strings.HasPrefix(c.config.Client.proxyURL.Scheme, "socks") {
+			// SOCKS5 proxy
+			if c.config.Client.proxyURL.Scheme != "socks" && c.config.Client.proxyURL.Scheme != "socks5h" {
```
breathbath, line 249: is too much nested maybe it's good to refactor it with early returns, dedicated functions etc

client/client.go
```diff
@@ -135,119 +131,206 @@ func (c *Client) connectionLoop(ctx context.Context) {
...
+
+	d := websocket.Dialer{
+		ReadBufferSize:   1024,
+		WriteBufferSize:  1024,
+		HandshakeTimeout: 45 * time.Second,
+		Subprotocols:     []string{chshare.ProtocolVersion},
+	}
+	//optionally proxy
+	if c.config.Client.proxyURL != nil {
+		if strings.HasPrefix(c.config.Client.proxyURL.Scheme, "socks") {
```
breathbath, line 247: a const for "socks"

client/client.go
```diff
@@ -135,119 +131,206 @@ func (c *Client) connectionLoop(ctx context.Context) {
...
+
+func (c *Client) connectToMainOrFallback() (conn *sshClientConn, isPrimary bool, err error) {
+	servers := append([]string{c.config.Client.Server}, c.config.Client.FallbackServers...)
+	for i, server := range servers {
+		conn, err = c.connect(server)
+		if err != nil {
+			c.Errorf(err.Error())
+			if _, ok := err.(retryableError); ok {
+				continue
+			} else {
```
breathbath, line 223: else not needed?

client/client.go
```diff
@@ -135,119 +131,206 @@ func (c *Client) connectionLoop(ctx context.Context) {
...
+	via := ""
+	if c.config.Client.proxyURL != nil {
+		via = " via " + c.config.Client.proxyURL.String()
+	}
+	c.Infof("Connecting to %s%s\n", server, via)
+
+	d := websocket.Dialer{
+		ReadBufferSize:   1024,
+		WriteBufferSize:  1024,
+		HandshakeTimeout: 45 * time.Second,
```
breathbath, line 242: a const would be good

client/client_test.go
```diff
@@ -286,3 +291,179 @@ func TestConnectionRequest(t *testing.T) {
...
+	assert.NoError(t, mainServer.WaitForStatus(true))
+
+	// connects to fallback if main server is down
+	mainServer.SetAvailable(false)
+	mainServer.CloseConnection()
+	assert.NoError(t, mainServer.WaitForStatus(false))
+	assert.NoError(t, fallbackServer.WaitForStatus(true))
+
+	// stays connected to fallback while main server id down
+	time.Sleep(200 * time.Millisecond)
```
breathbath, line 460: I am not sure if asserting against time is a good idea, e.g. if e.g. main server down takes more than 200ms or switching to fallback takes longer or the switch happens for 2ms and you don't need to wait the whole time.
Maybe it's a good idea to get some signals about server conn status changes and write tests to expect such signals (of course with a wait timeout if no expected signals were detected)

client/client_test.go
```diff
@@ -286,3 +291,179 @@ func TestConnectionRequest(t *testing.T) {
...
+	assert.NoError(t, mainServer.WaitForStatus(false))
+	assert.NoError(t, fallbackServer.WaitForStatus(true))
+
+	// connects back to main server when it becomes available
+	mainServer.SetAvailable(true)
+	time.Sleep(200 * time.Millisecond)
```
breathbath, line 466: same as above

client/config.go
```diff
@@ -28,16 +28,18 @@ type LogConfig struct {
...
-	Fingerprint     string        `mapstructure:"fingerprint"`
-	Auth            string        `mapstructure:"auth"`
-	Proxy           string        `mapstructure:"proxy"`
-	ID              string        `mapstructure:"id"`
-	Name            string        `mapstructure:"name"`
-	Tags            []string      `mapstructure:"tags"`
-	Remotes         []string      `mapstructure:"remotes"`
-	AllowRoot       bool          `mapstructure:"allow_root"`
-	UpdatesInterval time.Duration `mapstructure:"updates_interval"`
+	Server                   string        `mapstructure:"server"`
```
breathbath, line 31: I am just wondering if it would be maybe easier to give a sorted list of possible servers and rotate connections to them in order of appearance instead of defining main and fallback servers?

client/config.go
```diff
@@ -130,14 +135,35 @@ func (c *Config) parseServerURL() error {
 		return errors.New("server address is required")
 	}
 
+	url, err := c.parseURL(c.Client.Server)
```
breathbath, line 138: it repeats validation logic for the fallback server urls

client/config.go
```diff
@@ -130,14 +135,35 @@ func (c *Config) parseServerURL() error {
...
+		c.Client.FallbackServers[i] = url
+	}
+	return nil
+}
+
+func (Config) parseURL(urlStr string) (string, error) {
 	//apply default scheme
-	if !strings.Contains(c.Client.Server, "://") {
-		c.Client.Server = "http://" + c.Client.Server
+	if !strings.Contains(urlStr, "://") {
```
breathbath, line 160: what about `//server.com` in the urlStr which is valid but will be converted to non-valid `http:////server.com`?

client/client.go
domenipavec: Existing code.

client/client.go
domenipavec: Sure.

client/client.go
domenipavec: Existing code.

client/client_test.go
domenipavec: These sleeps are actually not needed, so I removed them.
I changed these tests to make them work so many times that the unnecessary sleeps stayed there. `WaitForStatus` will actually wait so no need for manual sleep.

client/client_test.go
domenipavec: Same

client/config.go
domenipavec: This is how the task has been defined. Also changing the server to slice would break backwards compatibility of config files.

client/config.go
domenipavec: Not sure what you mean, the parseURL is used to validate both server and fallback server urls and does not duplicate the code.

client/config.go
domenipavec: This is how it's implemented currently, I haven't made any changes to the validation/parsing logic. Feel free to file a bug for this if you think it's needed.

client/client.go
breathbath: ok

client/client.go
breathbath: ok

client/client.go
breathbath: no

client/client.go
breathbath: ok

client/client_test.go
breathbath: great!

client/client_test.go
breathbath: 👍

client/config.go
breathbath: I see ok

client/config.go
breathbath: ok if you say so

breathbath (approved):

PR #279, opened 2021-08-16
DEV-2273: Add single client get endpoint

domenipavec:
https://tracker.cloudradar.info/issue/DEV-2273


server/api.go
```diff
@@ -167,6 +167,7 @@ func (al *APIListener) initRouter() {
 	sub.HandleFunc("/me", al.handleChangeMe).Methods(http.MethodPut)
 	sub.HandleFunc("/me/ip", al.handleGetIP).Methods(http.MethodGet)
 	sub.HandleFunc("/clients", al.handleGetClients).Methods(http.MethodGet)
+	sub.HandleFunc("/clients/{client_id}", al.wrapClientAccessMiddleware(al.handleGetClient)).Methods(http.MethodGet)
```
m-terel, line 170: Please add API docs for the new endpoint.

server/api.go
```diff
@@ -614,6 +615,20 @@ func (al *APIListener) handleGetClients(w http.ResponseWriter, req *http.Request
...
+	vars := mux.Vars(req)
+	clientID := vars[routeParamClientID]
+
+	client, err := al.clientService.GetByID(clientID)
+	if err != nil {
+		al.jsonError(w, err)
+		return
+	}
+
+	clientPayload := convertToClientPayload(client)
```
m-terel, line 628: Currently, this can cause nil pointer. Please handle a case when a client not found. Return 404.

server/api_test.go
```diff
@@ -1546,3 +1546,87 @@ func TestHandleRefreshUpdatesStatus(t *testing.T) {
...
+			},
+		},
+	}
+	al.initRouter()
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest("GET", "/api/v1/clients/client-1", nil)
+	al.router.ServeHTTP(w, req)
+
+	expectedJSON := `{
```
m-terel, line 1567: It could be difficult to maintain such a test because if the client struct changes it would always require updating this test. It might be better to convert the `w.Body.String()` to client struct and compare values with `c1`. Or to have a func to get the json string from the client struct.

server/api.go
domenipavec: Forgot about that, done

server/api.go
domenipavec: I haven't detected that because authentication middleware already rejects the request if client doesn't exist with 404. But I guess it's better to not rely on that, so I added it.

server/api_test.go
domenipavec: I find it harder to maintain tests that compare specific values, since the test doesn't fail when adding new values, so inevitably some values don't get tested. I think it's better if the test fails if the output changes for any reason so we know about it, it's not that hard to change this value when needed.
Also the test for listing clients already uses the same test pattern.

m-terel (approved):

PR #282, opened 2021-08-22
Autoclose tunnels after 5 minutes by default

breathbath:


api-doc.yml
```diff
@@ -446,11 +446,17 @@ paths:
         type: "string"
       - name: "idle-timeout-minutes"
         in: "query"
-        description: "Auto-close the tunnel after given period of inactivity in minutes. If 0 or not provided - auto-closing is disabled."
+        description: "Auto-close the tunnel after given period of inactivity in minutes. If no provided, default value is 5 minutes. This parameter should not be used with a non empty `skip-idle-timeout` parameter"
```
domenipavec, line 449:
```suggestion
        description: "Auto-close the tunnel after given period of inactivity in minutes. If not provided, default value is 5 minutes. This parameter should not be used with a non empty `skip-idle-timeout` parameter"
```

api-doc.yml
```diff
@@ -446,11 +446,17 @@ paths:
...
         required: false
         type: "integer"
         maximum: 10080
         minimum: 0
+        default: 5
+      - name: "skip-idle-timeout"
+        in: "query"
+        description: "Disables the auto-close time of a creatable tunnel (see `idle-timeout-minutes` parameter). The parameter should not be used with a non empty `idle-timeout-minutes` parameter"
```
domenipavec, line 457:
```suggestion
        description: "Disables the auto-close time of the tunnel (see `idle-timeout-minutes` parameter). The parameter should not be used with a non empty `idle-timeout-minutes` parameter"
```
Not sure what `a creatable` is supposed to mean, but it seems it's referred to as `the tunnel` in all other params.

server/api.go
```diff
@@ -264,23 +264,26 @@ func (al *APIListener) writeJSONResponse(w http.ResponseWriter, statusCode int,
 }
 
 func (al *APIListener) jsonErrorResponse(w http.ResponseWriter, statusCode int, err error) {
-	al.writeJSONResponse(w, statusCode, api.NewErrorPayload(err))
+	al.writeJSONResponse(w, statusCode, api.NewErrorPayload(err, "", ""))
```
domenipavec, line 267: Nit.: It seems `api.NewErrorPayload` and `api.NewErrorPayloadWithCode` are very similar now, would it make sense to merge them together?

server/api.go
```diff
@@ -858,18 +861,67 @@ const (
...
+	if idleTimeoutMinutesStr != "" && skipIdleTimeout != "" {
+		return 0, errors2.APIError{
+			Message: fmt.Sprintf(
+				"conflicting parameters: %s=%s and %s=%s, either decide for skipping an idle timeout or provide a non empty timeout value",
+				idleTimeoutMinutesQueryParam,
+				idleTimeoutMinutesStr,
+				skipIdleTimeoutQueryParam,
+				idleTimeoutMinutesStr,
+			),
+			Code:    http.StatusBadRequest,
```
domenipavec, line 891: It's really unclear having both `Code` and `ErrCode`. It's hard to make sense of what is what, especially as `ErrCode` ends up in `Code` field of `ErrorPayloadItem`. Maybe rename this field to `Status` and the other to `Code`?

server/api.go
```diff
@@ -858,18 +861,67 @@ const (
...
+	ErrCodeRemotePortNotOpen      = "ERR_CODE_REMOTE_PORT_NOT_OPEN"
+	ErrCodeTunnelExist            = "ERR_CODE_TUNNEL_EXIST"
+	ErrCodeTunnelToPortExist      = "ERR_CODE_TUNNEL_TO_PORT_EXIST"
+	ErrCodeURISchemeLengthExceed  = "ERR_CODE_URI_SCHEME_LENGTH_EXCEED"
+	ErrCodeInvalidACL             = "ERR_CODE_INVALID_ACL"
+	ErrCodeInvalidIdleTimeout     = "ERR_CODE_INVALID_IDLE_TIMEOUT"
+	ErrCodeConflictingIdleTimeout = "ERR_CODE_IDLE_TIMEOUT_PARAMETER_CONFLICT"
 )
 
+func (al *APIListener) resolveIdleTunnelTimeoutValue(req *http.Request) (int, error) {
```
domenipavec, line 879: Nit.: Maybe pass query instead of the whole request since the whole request is not needed?
Also since this method does not require anything from `APIListener` maybe it would be nice for it to be an independent function? Maybe even move it to separate `validation` file or package.

server/api.go
```diff
@@ -858,18 +861,67 @@ const (
...
 	idleTimeoutMax               = 7 * 24 * 60 // week
-
-	ErrCodeLocalPortInUse        = "ERR_CODE_LOCAL_PORT_IN_USE"
-	ErrCodeRemotePortNotOpen     = "ERR_CODE_REMOTE_PORT_NOT_OPEN"
-	ErrCodeTunnelExist           = "ERR_CODE_TUNNEL_EXIST"
-	ErrCodeTunnelToPortExist     = "ERR_CODE_TUNNEL_TO_PORT_EXIST"
-	ErrCodeURISchemeLengthExceed = "ERR_CODE_URI_SCHEME_LENGTH_EXCEED"
-	ErrCodeInvalidACL            = "ERR_CODE_INVALID_ACL"
-	ErrCodeInvalidIdleTimeout    = "ERR_CODE_INVALID_IDLE_TIMEOUT"
+	idleTimeoutDefault           = 5
```
domenipavec, line 867: Add a comment that these are minutes?

server/api.go
```diff
@@ -858,18 +861,67 @@ const (
...
+				idleTimeoutMinutesQueryParam,
+				idleTimeoutMinutesStr,
+				skipIdleTimeoutQueryParam,
+				idleTimeoutMinutesStr,
+			),
+			Code:    http.StatusBadRequest,
+			ErrCode: ErrCodeConflictingIdleTimeout,
+		}
+	}
+	if skipIdleTimeout != "" {
```
domenipavec, line 895: The behaviour of `skipIdleTimeout` is just a bit weird, I would expect to be able to set it to `true` or `false`. Not just any string that is set is `true`.
I think it would make sense to use `strconv.ParseBool` here.

api-doc.yml
breathbath: ok

api-doc.yml
breathbath: ok

server/api.go
breathbath: In the name they seem to be similar but in the implementation they are quite different (one works with plain strings and another with error interface) and merge them into 1 method would introduce a complex method where some parameters should be handled with defaults. My proposal is to rename both to make the difference more distinct

server/api.go
breathbath: ok. I think the problem is in the legacy code, we try to make it consistent and at the end we figure out that error codes are actually not really used/needed. I would completely discard them and use http codes but I cannot make this decision

server/api.go
breathbath: ok

server/api.go
breathbath: fixed

server/api.go
breathbath: ok

server/api.go
```diff
@@ -901,20 +904,18 @@ func (al *APIListener) handlePutClientTunnel(w http.ResponseWriter, req *http.Re
...
 
 	idleTimeoutMinutesStr := req.URL.Query().Get(idleTimeoutMinutesQueryParam)
-	var idleTimeoutMinutes int
-	if idleTimeoutMinutesStr != "" {
-		idleTimeoutMinutes, err = strconv.Atoi(idleTimeoutMinutesStr)
-		if err != nil {
-			al.jsonErrorResponseWithError(w, http.StatusBadRequest, ErrCodeInvalidIdleTimeout, fmt.Sprintf("invalid %q param", idleTimeoutMinutesQueryParam), err)
-			return
-		}
+	skipIdleTimeout, err := strconv.ParseBool(req.URL.Query().Get(skipIdleTimeoutQueryParam))
```
domenipavec, line 907: Nit.: It would be better to be consistent where you're parsing these query arguments. Either both here or both in `ResolveIdleTunnelTimeoutValue`.

domenipavec (approved):

PR #288, opened 2021-09-01
Taco as interpreter + Cmd buf overflow error

breathbath:


client/exec.go
```diff
@@ -216,6 +217,10 @@ var getInterpreter = func(inputInterpreter, os string, hasShebang bool) (string,
 		return "", nil
 	}
 
+	if inputInterpreter == taco {
```
domenipavec, line 220: This will only be available on linux, since windows validation returns above. Shouldn't taco be available on windows as well?
Also it would be nice to have some tests for this.

server/script/executor.go
```diff
@@ -74,7 +76,12 @@ func (e *Executor) createClientScriptPath(os, interpreter string) (string, error
 		return scriptName + ".bat", nil
 	}
 
-	return scriptName + ".sh", nil
+	extension := ".sh"
+	if interpreter == validation.Taco {
+		extension = ".yml"
```
domenipavec, line 81: This will only set extension for linux, on windows taco scripts will end up with `.bat` extension.

server/validation/interpreter.go
```diff
@@ -0,0 +1,27 @@
+package validation
+
+import "fmt"
+
+const Taco = "taco"
```
domenipavec, line 5: It seems we're duplicating the constants from `client/exec.go`, it would be better to have them defined in `share` I think.

domenipavec (commented):
Would be nice to have some tests for new code.

client/exec.go
breathbath: ok

server/script/executor.go
breathbath: ok

server/validation/interpreter.go
breathbath: ok

server/script/executor_test.go
```diff
@@ -4,59 +4,116 @@ import (
...
+	const defaultHash = "a1159e9df3670d549d04524532629f5477ceb7deec9b45e47e8c009506ecb2c8" //sha256 of "pwd"
+	const defaultScript = "pwd"
+	testCases := []struct {
+		name                  string
+		filePathToReturn      string
+		Sha256HashToReturn    string
+		clientOSKernelToGive  string
+		scriptToGive          string
+		fileExtensionToExpect string
+		interpreterToGive     string
```
domenipavec, line 29: This naming with `ToReturn` and `ToGive` suffixes is not a common practice and confused me. It's common to use `want` or `expected` prefix for anything that is asserted against in test and no prefix for anything that is used as an argument.

breathbath: sorry but I don't agree about this common

domenipavec: Here's a few examples from quick google search, they're all using `want`:
https://golang.org/doc/tutorial/add-a-test
https://gobyexample.com/testing
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
Also in our code:
https://github.com/cloudradar-monitoring/rport/blob/dc592468b4cd9fdb95effaaae6e76a999aae4130/client/exec_test.go#L124

breathbath: The given word is quite common in integration test frameworks e.g. 
https://www.cloudbees.com/blog/implementing-a-bdd-workflow-in-go
https://github.com/cucumber/godog
but honestly we're wasting time if we discuss such small things like how we name properties in mocks
I'd better invest this time to refactor rport code to get rid of api.go monster and do a proper integration and unittesting for parts that are not covered

breathbath: anyway if you insist I can refactor this part

domenipavec: If the test was clear to understand, then I wouldn't care that much about specific words. But this test is really hard to understand anyway so the words I wasn't used to didn't really help (I would still prefer to have `want` as prefix instead of `suffix`, but that's more of a nit-pick).
Here's what I would to make the test easier to understand:
1. remove hash and script from test cases since they're the same for all.
2. put interpreter and clientOS first as they're inputs and outputs filePath and fileExtension last as `want`.
3. This `CreateScriptOnClient` method could really use a refactor since it's doing 3 things (checking hash, creating a filename and communicating with the client).
But I'll approve for now so this doesn't drag on forever.

domenipavec (approved):

server/script/executor_test.go
breathbath: thanks

breathbath: yes we could argue further but anyway let's be pragmatic and objective

domenipavec (approved):

PR #295, opened 2021-09-14
Cmd buf overflow error

breathbath:


client/exec.go
```diff
@@ -252,22 +271,33 @@ func matchRegexp(cmd string, regexpList []*regexp.Regexp) bool {
...
 
 	// do not write to buffer if no space left
-	if freeCapacity <= 0 {
-		return len(p), nil // pretend a successful write
-	}
-
-	// write to buffer only a part if exceeds the capacity
 	if len(p) > freeCapacity {
-		return b.Buffer.Write(p[:freeCapacity])
+		b.hasOverflow = true
```
domenipavec, line 292: Since you're not writing in this branch you're potentially losing data that could still be sent back: if `freeCapacity` is 500 bytes and `len(p)` is 512 bytes, then you could write and send to the user the first 500 bytes and then stop.

client/exec.go
```diff
@@ -252,22 +271,33 @@ func matchRegexp(cmd string, regexpList []*regexp.Regexp) bool {
...
+		return 0, errors.New(b.GetOverflowMessage())
```
domenipavec, line 293: Are you sure returning an error here is a good idea? Previous implementation would just silently drop the overflow data and pretend everything is ok for the exec package.
Have you tested what the `exec` package does when error happens writing to stderr or stdout? Does it stop the execution of the command or continues? Does it report the error somewhere?

breathbath: it was the requirement of the task - return error is buffer is overflown 
I've tested it the exec returns "broken pipe error" so I have to catch it in a different way (through the flag)
yes it does stop the execution with an stderr

client/exec.go
breathbath: I didn't have exactly this requirement in the ticket but I can do it

breathbath: done

domenipavec (approved):

PR #298, opened 2021-09-20
DEV-2316 Fixing space escaping

breathbath:


client/exec_nix.go
```diff
@@ -19,7 +19,11 @@ func (e *CmdExecutorImpl) New(ctx context.Context, execCtx *CmdExecutorContext)
 		args = append(args, interpreter, "-c")
 	}
 
-	commandStr := strings.ReplaceAll(execCtx.Command, " ", "\\ ")
+	commandStr := execCtx.Command
+	if execCtx.IsScript {
+		commandStr = strings.ReplaceAll(commandStr, " ", "\\ ")
```
domenipavec, line 24: I'm not sure what the use case for this is and in what cases it should be escaped and when not. Can you add tests?

breathbath: it should be escaped in case when a script is executed

breathbath: unfortunately this is an urgent blocker as without this fix all commands with spaces are not executable

PR #332, opened 2021-10-25
Dev 2370 auth app 2fa

breathbath:


devgek (dismissed):

cmd/rportd/main.go
```diff
@@ -302,6 +304,8 @@ func init() {
 	viperCfg.SetDefault("api.two_fa_send_to_type", message.ValidationNone)
 	viperCfg.SetDefault("api.enable_audit_log", true)
 	viperCfg.SetDefault("api.audit_log_rotation", auditlog.RotationMonthly)
+	viperCfg.SetDefault("api.totp_secret", "")
```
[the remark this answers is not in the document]
breathbath: ok

devgek (approved):

server/bearer_auth.go
```diff
@@ -67,11 +72,28 @@ func (al *APIListener) validateBearerToken(tokenStr string) (bool, string, *APIS
...
+		al.Errorf(
+			"Token subject %s doesn't match the current url %s, so this token is not intended to be used for this page",
+			tk.StandardClaims.Subject,
+			curSubject,
+		)
+		return false, tk.Username, nil, nil
+	}
+
 	if al.bannedUsers.IsBanned(tk.Username) {
+		al.Errorf(
```
domenipavec, line 85: Not sure all these error logs are the best idea. Especially this one - if someone is trying to brute force the password and get banned, they could really increase the log size by hitting this log point.

server/api.go
```diff
@@ -755,6 +792,32 @@ func (al *APIListener) handleDeleteUser(w http.ResponseWriter, req *http.Request
...
+	if err != nil {
+		al.jsonError(w, err)
+		return
+	}
+	if user == nil {
+		al.jsonErrorResponseWithTitle(w, http.StatusNotFound, "user not found")
+		return
+	}
+
+	al.auditLog.Entry("users-totp-secret", auditlog.ActionDelete).
```
domenipavec, line 813: Please add the application to auditlog constants, so we have a central list of all possible values.
Also add saving auditlog entries to other endpoints that modify totp secrets.

server/api.go
```diff
@@ -1173,6 +1236,92 @@ func (al *APIListener) handleGetMe(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	if err := al.userService.Change(userDataToChange, user.Username); err != nil {
+		al.jsonError(w, err)
+		return
+	}
+
+	if action == "create" {
+		al.Debugf("Users time based one time secret is created for user [%s].", user.Username)
+		w.WriteHeader(http.StatusCreated)
```
domenipavec, line 1318: It would be nice to return the secret and image to user on create request, so they don't have to make another request.

server/totp.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func StoreTotPCodeInUser(usr *users.User, totP *TotP) error {
+	totPBytes, err := json.Marshal(totP)
+
+	if err != nil {
+		return fmt.Errorf("failed to generate totP secret data: %v", err)
+	}
+
+	usr.TotP = string(totPBytes)
```
domenipavec, line 53: Not sure how big the image is, but it's a bit weird we're saving that in db. We could just generate it on the fly when needed.

server/api.go
```diff
@@ -1173,6 +1236,92 @@ func (al *APIListener) handleGetMe(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		al.Logf(chshare.LogLevelError, "failed to get TotP secret: %v", err)
+		al.jsonErrorResponse(w, http.StatusInternalServerError, err)
+		return
+	}
+	al.writeJSONResponse(w, http.StatusOK, totP)
+}
+
+func (al *APIListener) handlePostTotP(w http.ResponseWriter, req *http.Request) {
+	issuer := req.URL.Query().Get("application-name")
```
domenipavec, line 1261: Not sure what the use case would be to have the user set the issuer themselves. I think the issuer should just be a constant.

server/api.go
```diff
@@ -1173,6 +1236,92 @@ func (al *APIListener) handleGetMe(w http.ResponseWriter, req *http.Request) {
 	al.writeJSONResponse(w, http.StatusOK, response)
 }
 
+func (al *APIListener) handleGetTotP(w http.ResponseWriter, req *http.Request) {
```
domenipavec, line 1239: I'm not sure if it has any security implications to allow users to read their secret later, but most sites I've seen only show you the code when it's first generated and don't allow reading it again. So I would drop the get endpoint here and only return the code to the user in create endpoint.

server/config.go
```diff
@@ -49,7 +49,9 @@ type APIConfig struct {
 	TwoFASendToRegex         string                 `mapstructure:"two_fa_send_to_regex"`
 	twoFASendToRegexCompiled *regexp.Regexp
 
-	AuditLog auditlog.Config `mapstructure:",squash"`
+	AuditLog                auditlog.Config `mapstructure:",squash"`
+	TotPEnabled             bool            `mapstructure:"totp_enabled"`
```
domenipavec, line 53: Since you're treating this differently from 2FA, you have to check this as well anywhere where `IsTwoFAOn` is checked. For example in `handleBasicAuth`, where it's decided whether username and password should be accepted. Otherwise using basic auth would be possible even when totp is enabled. 
Might be easier to add it to `IsTwoFAOn`?

server/api.go
```diff
@@ -406,17 +411,35 @@ func (al *APIListener) handleLogin(username, pwd string, w http.ResponseWriter,
...
 	if err != nil {
 		al.jsonErrorResponse(w, http.StatusBadRequest, err)
 		return
 	}
 
-	tokenStr, err := al.createAuthToken(lifetime, username)
+	if al.config.API.TotPEnabled {
+		al.twoFASrv.SetTotPLoginSession(username, al.config.API.TotPLoginSessionTimeout)
+
+		tokenStr, err := al.createAuthToken(lifetime, username, "/api/v1/me/totp-secret")
```
domenipavec, line 423: You definitely shouldn't always give access to `/me/totp-secret` - it would be ok on the first login when user doesn't have the secret so they can generate it. But for later logins, this exposes a huge security hole, since the user can just read the secret and generate the token - basically bypassing totp 2FA protection.

server/api.go
```diff
@@ -580,6 +609,14 @@ func (al *APIListener) parseAndValidate2FATokenRequest(req *http.Request) (usern
 		}
 	}
 
+	if al.config.API.TotPEnabled {
```
domenipavec, line 612: You should verify the jwt token you're returning after login in this endpoint as well to solve the 2nd vulnerability I described in my previous review:
```Requiring no password or server generated token when validating 2fa, means that it's easy to login without knowing the password with only totp token. I simply try every few minutes, eventually the user will try to login, and I will also be able to login as that user with only totp token.```

**Review by domenipavec — CHANGES_REQUESTED** [review]

File: rportd.example.conf [hunk 11]
```diff
@@ -273,6 +273,15 @@
   ## Consider changing to a faster rotation.
   #audit_log_rotation = 'monthly' , possible values: yearly, monthly, weekly, daily
 
+  ## To enable time-based onetime token generated by app likes Google or Microsoft Authenticator,
+  ## totp_enabled = true.
+  ## Your user-password store (json files or DB table) needs an additional text field 'totp_secret'.
+  # totp_enabled = false
```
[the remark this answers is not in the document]
breathbath: ok

server/bearer_auth.go
breathbath: ok

server/api.go
breathbath: ok done

server/api.go
breathbath: ok

server/totp.go
breathbath: hey it was tickets requirement, but the idea is great I changed the implementation to the runtime generation

server/api.go
breathbath: The use case is that this field is visible in the authenticator app as the application name, maybe someone is unhappy about the "rport" name and wants to use another name e.g. "super-fancy-server-management-tool"

server/api.go
breathbath: this was a ticket requirement, but anyway if you can generate and read it at any time it's the same security problem as to just read it. The only security issue here is that we store the secret in database which might be exposed. But as mentioned - this is one of the requirements.

server/config.go
breathbath: I've added this check to the basic auth use case, but I would not add it as part of `IsTwoFAOn` as it might interfere with other use cases where IsTwoFAOn is expected to be true but TotPEnabled is false

server/api.go
breathbath: first it's a ticket requirement, secondly there is no hole because to use this api you should first pass login an generate a special jwt token which cannot be used in other APIs.

server/api.go
breathbath: sorry I would not use the word "easy" here: imagine you have to guess a 6 digit number which is changing every 30s and do it in the time range between eventual login and user's 2fa auth success or login attempt timeout. But actually the same issue is with the existing 2fa, I might try to guess the code which was sent to the user in the time between login and 2fa check. 

Our current implementation  assumes that the user who tried to login is the same who is trying the 2fa code by the fact that user provided a correct code in the expected time range, and there we could also use some intermediate jwt token to relate both actions.

server/api.go
domenipavec: I guess, don't really care about this that much :)

server/api.go
domenipavec: Ticket requirements clearly state that access to `/me/totp-secret`  should be given with only username and password only if user hasn't generate the secret yet:

> If totp_secret IS NULL we must permit the log in with just username and password and the only Endpoint, the user is allowed to use is /me/totp-secret. This enables and forces the user to create the secrete and activate the app.

Once the user has generated the secret allowing them access to `/me/totp-secret` without the 2FA token bypasses the need for having the ability to generate totp tokens.

For example I find out someone's username and password. I'm able to login with that username and password and access `/me/totp-secret` to generate a new totp secret with which I can then gain full access.

The point of 2FA/totp authentication is that anyone logging in must know both password and the token to gain access.

server/api.go
domenipavec: The point of 2FA/totp authentication is that anyone logging in must know both password and the token to gain access.

If I somehow gain access to the totp secret I'm able to generate totp token I still must not be able to login without the password. Which your implementation enables me: if I just keep trying to login I will eventually be able to login once the user has entered password. This is not about guessing the token but about the case where I know the totp secret.

You should not assume that the user who provided the password is the same as the one providing the token. As you mentioned you should use the intermediate jwt to connect the actions.

server/api.go
breathbath: ???

server/api.go
breathbath: if user looses access to his phone he automatically looses access to his account.  I understand /me/totp-secret as a way to do this and also do the first validation, e.g. with Google 2fa I can manage totp secret keys at any time, but probably after the second check and possibly another password request, but the ticket doesn't mention this case at all.

server/api.go
breathbath: same applies for the current 2fa - a hacker gains access to the users email + 2fa code and applies it without knowing login and password.

server/api.go
breathbath: token post api can only be used with the login token for the first time, in all other cases we require a token which is given after the totp check

server/api.go
breathbath: I changed the code as follows:
- if totp is enabled, user gets a special login token after login api (a token with totp scope)
- the login token can used to create a new totp secret (POST to /me/tot-p-secret) only if no other secret was created
- the login token is required in /check-2fa if totp is enabled, if a valid totp code is given, user gets a proper token (with an empty scope)
- the login token cannot be used in all other apis
- the proper token can be used to create, delete and read totp secrets

server/api.go
breathbath: I've solved it see my previous comment

api-doc.yml
```diff
@@ -76,7 +76,9 @@ paths:
       summary: "Generate auth token. Requires username and password in request body"
       description: "If 2FA is disabled: returns authorization JWT token.\n
       If 2FA is enabled: triggers sending 2FA token to the user. 2fa token can be further verified using `/verify-2fa` endpoint. Requires the user to have a valid `two_fa_send_to` delivery option.\n
-      `username` and `password` should be provided in request body either in JSON either in x-www-formurlencoded format"
+      `username` and `password` should be provided in request body either in JSON either in x-www-formurlencoded format.\n
+      If time based one time passwords (TotP) are enabled (Google/Microsoft authenticator app), returns authorization JWT token. However this token can be used only for `/me/totp-secret` URLs.\n
```
domenipavec, line 80: You should update the description.

server/2fa_auth.go
```diff
@@ -99,6 +101,61 @@ func (srv *TwoFAService) SendToken(ctx context.Context, username string) (sendTo
 	return user.TwoFASendTo, nil
 }
 
+func (srv *TwoFAService) SetTotPLoginSession(username string, loginSessionTTL time.Duration) {
+	srv.mu.Lock()
+	srv.tokensByUser[username] = &expirableToken{
+		expiry: time.Now().Add(loginSessionTTL),
```
domenipavec, line 107: Nit.: I don't think tracking expiry here is necessary anymore, since you generate jwt token with expiry on login anyway.

server/2fa_auth_test.go
```diff
@@ -0,0 +1,111 @@
...
+		Issuer:      "iss2",
+		AccountName: "acc2",
+	}
+	totP, err := GenerateTotPSecretKey(inpt)
+	require.NoError(t, err)
+
+	StoreTotPCodeInUser(usr, totP)
+
+	err = tfaService.ValidateTotPCode(usr, "dfasdf")
+	require.EqualError(t, err, "invalid codem")
```
domenipavec, line 63: Typo?

server/api.go
```diff
@@ -533,21 +611,23 @@ func (al *APIListener) handleDeleteLogout(w http.ResponseWriter, req *http.Reque
 	w.WriteHeader(http.StatusNoContent)
 }
 
-func (al *APIListener) handlePostVerify2FAToken(w http.ResponseWriter, req *http.Request) {
-	username, err := al.parseAndValidate2FATokenRequest(req)
-	if err != nil {
-		if !al.handleBannedIPs(w, req, false) {
+func (al *APIListener) handlePostVerify2FAToken() http.Handler {
```
domenipavec, line 614: Why is this method returning a handler instead of being the handler itself? It doesn't seem you're using closure so it would be required?

server/bearer_auth.go
```diff
@@ -20,11 +20,49 @@ const (
...
 	jwt.StandardClaims
 }
 
-func (al *APIListener) createAuthToken(ctx context.Context, lifetime time.Duration, username string) (string, error) {
+type Scope struct {
+	URI    string `json:"uri,omitempty"`
+	Method string `json:"method,omitempty"`
+}
+
+var ScopesExcluding2FaCheck = []Scope{
```
domenipavec, line 33: I think the name is confusing. Shouldn't it be like `Allow all` scope or something?

domenipavec (approved):
A few minor comments, otherwise LGTM.

PR #350, opened 2021-11-05
Dev 2343 monitoring api

devgek:


api-doc.yml
```diff
@@ -2208,6 +2208,155 @@ paths:
...
+          description: "There is only `timestamp` allowed as sort field. Default direction is DESC\n
+            To sort ascending use `&sort=-timestamp`."
+          required: false
+          type: "string"
+        - in: "query"
+          name: "filter"
+          description: "Filter to find measurement points. It must be a combination of lower- and upper-bound.\n
+              It should be provided in the format as `filter[<FIELD>][<OP-LOWER>]=<VALUE>&filter[<FIELD>][<OP-UPPER>]=<VALUE>`,\n
+              where `<FIELD>` is one of the values `timestamp` or `limit`, `OP-LOWER` is one of `gt` or `since`, `OP-UPPER` is one of `lt` or `until`, and `<VALUE>` is the search value,\n
+              e.g. `filter[timestamp][gt]=1636009200&filter[timestamp[lt]=1636009500` will request measurement points between the given timestamps(`unixepoch`).\n
```
domenipavec, line 2236: Typo, missing `]` after timestamp and before `[lt]`. Same in the next line.

server/monitoring/service.go
```diff
@@ -16,20 +16,21 @@ import (
...
+	GetClientMetricsLatest(ctx context.Context, clientID string, o *query.ListOptions) (*monitoring.ClientMetricsPayload, error)
+	GetClientMetricsFiltered(ctx context.Context, clientID string, o *query.ListOptions) ([]monitoring.ClientMetricsPayload, error)
 	GetClientProcessesLatest(ctx context.Context, clientID string) (*monitoring.ClientProcessesPayload, error)
 	GetClientProcessesFiltered(ctx context.Context, clientID string, filters []query.FilterOption) (*monitoring.ClientProcessesPayload, error)
 	GetClientMountpointsLatest(ctx context.Context, clientID string) (*monitoring.ClientMountpointsPayload, error)
 	GetClientMountpointsFiltered(ctx context.Context, clientID string, filters []query.FilterOption) (*monitoring.ClientMountpointsPayload, error)
 }
 
-var layoutSinceUntil = "2006-01-02:15:04:05"
+var layoutAPI = "2006-01-02:15:04:05"
```
domenipavec, line 27: This is a weird format to use, why not use something standard like `time.RFC3339`?

share/query/convert.go
```diff
@@ -23,6 +23,7 @@ func AppendOptionsToQuery(o *ListOptions, q string, inParams []interface{}) (str
 	qOut, params := addWhere(o.Filters, q)
 	outParams := append(inParams, params...)
 	qOut = addOrderBy(o.Sorts, qOut)
+	qOut = addLimit(o.Filters, qOut)
```
domenipavec, line 26: I've added `Pagination` option to query recently for limit and offset. It would be nice if you use that, so our api is consistent. Instead of having multiple ways to specify limit.

db/migration/monitoring/bindata.go
```diff
@@ -84,12 +84,12 @@ func _001_initDownSql() (*asset, error) {
...
+var __001_initUpSql = []byte("\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\xff\x84\xd0\xcd\x4a\x03\x31\x14\x05\xe0\x7d\x9e\xe2\x30\xab\x16\x9a\x27\x70\x15\x35\xc2\xe0\xb4\xca\xf4\x0a\xed\x6a\x18\xe3\x55\x02\xcd\x24\xe4\x67\xe1\xdb\x4b\xa9\x2d\x23\x1d\xa7\x77\x95\x45\xbe\x73\x93\x23\x25\xe4\xcc\x08\x29\x41\xfd\xfb\x81\x91\x72\x2c\x26\x97\xc8\xf8\xf4\x11\x8e\xfb\x54\x22\x3b\x1e\x72\x12\xb7\x32\x1e\x5a\xad\x48\x83\xd4\x7d\xa3\x51\x3f\x61\xf3\x42\xd0\xbb\x7a\x4b\x5b\x54\xe3\xa0\x4a\x2c\x04\x00\x54\xe6\x60\x79\xc8\x9d\xfd\xa8\x30\x1e\xd2\x3b\x3a\x9f\x8f\x19\x9b\xb7\xa6\x59\x9d\x44\xb6\x8e\x53\xee\x5d\xf8\x2b\x1e\x15\x69\xaa\xd7\x7a\x2c\xf0\x4b\x4c\x28\x5d\x49\xfd\x17\x77\x81\xa3\xe1\x21\x9f\x68\xab\x55\xf3\xcf\x12\xc7\xce\xc7\xef\x2b\x34\x23\xac\x9f\x5a\x31\x27\x42\xf4\x86\x53\xe2\x74\xfd\xf5\xf3\x2b\x7c\x19\x72\xf0\xf6\x58\xd8\xe4\x8d\xd7\xb6\x5e\xab\x76\x8f\x67\xbd\xc7\xe2\x52\xe5\x0a\x97\x8e\x96\x62\x79\x27\x7e\x02\x00\x00\xff\xff\xce\xe5\xad\x53\xf9\x01\x00\x00")
```
domenipavec, line 92: You probably shouldn't change this directly without also changing the source `sql` file. 
Also you shouldn't be changing the migration, what about the users that are running code that has already run this migration? They won't run the updated migration and have inconsistent schema. If you made changes you should write a new migration that applies those changes.

server/api.go
```diff
@@ -3191,15 +3191,15 @@ func (al *APIListener) handleGetClientMetrics(w http.ResponseWriter, req *http.R
 	}
 
 	if queryOptions.HasFilters() {
-		al.handleGetClientMetricsList(w, req, clientID, queryOptions)
+		al.handleGetClientMetricsFiltered(w, req, clientID, queryOptions)
 		return
 	}
 
-	al.handleGetClientMetricsOne(w, req, clientID, queryOptions)
+	al.handleGetClientMetricsLatest(w, req, clientID, queryOptions)
```
domenipavec, line 3198: This api implementation is a bit weird that you return only 1 result with no filters. What I would expect from list endpoints would be to have a default limit and it would return that many results.

server/api/monitoring/payload.go
```diff
@@ -1,26 +1,44 @@
 package monitoring
 
+import (
+	"time"
+
+	"github.com/cloudradar-monitoring/rport/share/types"
+)
+
+type CPUUsagePercent struct {
+	Value float64 `json:"value,omitempty" db:"cpu_usage_percent"`
```
domenipavec, line 10: Having this structure and `percent` in the name even when you're returning actual values and not aggregations is a bit weird. It also increases the size of the response significantly to have another object in response instead of direct value.

server/monitoring/service.go
```diff
@@ -16,20 +16,21 @@ import (
...
+var layoutDb = "2006-01-02 15:04:05"
+var maxLimit = 120
```
domenipavec, line 29: You could make these `const` instead of var.

server/monitoring/service.go
```diff
@@ -16,20 +16,21 @@ import (
...
 var maxDataFetchHours = 48
 var maxDataFetchDuration = time.Duration(maxDataFetchHours) * time.Hour
-
-//var limitDownSamplingHours = 2
+var thresholdDownsamplingHours = 2
+var thresholdDownsamplingDuration = time.Duration(thresholdDownsamplingHours) * time.Hour
```
domenipavec, line 33: No need to have 2 variables if you're not using the `thresholdDownsamplingHours` directly anywhere. Just set it to `2 * time.Hour` directly. It would also be more readable that way.
Same for the one above.

server/monitoring/service.go
```diff
@@ -47,18 +48,18 @@ func (s *monitoringService) DeleteMeasurementsOlderThan(ctx context.Context, per
...
-}
-
 func (s *monitoringService) GetClientProcessesLatest(ctx context.Context, clientID string) (*monitoring.ClientProcessesPayload, error) {
 	return s.DBProvider.GetProcessesLatestByClientID(ctx, clientID)
 }
 
 func (s *monitoringService) GetClientProcessesFiltered(ctx context.Context, clientID string, filters []query.FilterOption) (*monitoring.ClientProcessesPayload, error) {
-	t, err := time.Parse(layoutSinceUntil, filters[0].Values[0])
+	var t time.Time
+	_, err := strconv.Atoi(filters[0].Values[0])
```
domenipavec, line 57: This code is super weird, I'm assuming that filter can be an int sometimes, and you want to parse to time just when it's not?
It would be better to use column names to determine conversion instead of this brute force trying.

server/monitoring/service.go
```diff
@@ -68,77 +69,107 @@ func (s *monitoringService) GetClientMountpointsLatest(ctx context.Context, clie
...
+		return nil, errors.APIError{Message: fmt.Sprintf("Illegal limit (allowed: 1 to %d)", maxLimit), HTTPStatus: http.StatusBadRequest}
+	}
+	return s.DBProvider.GetMetricsListByClientID(ctx, clientID, o)
+}
+
+func (s *monitoringService) getClientMetricsFilteredRange(ctx context.Context, clientID string, o *query.ListOptions) ([]monitoring.ClientMetricsPayload, error) {
+	lower, _ := time.Parse(layoutDb, o.Filters[0].Values[0])
+	upper, _ := time.Parse(layoutDb, o.Filters[1].Values[0])
+
+	if upper.Before(lower) {
```
domenipavec, line 113: This deciding on what to return based on filtered range is really weird and confusing for api users.
Here's how I would implement this:
Allow any timestamp filters. Then use `Pagination` with limit and offset to limit the number of results. This is how I implemented `auditlog` endpoint if you want a reference.
Make a separate endpoint for downsampled results, so it's easier for api users to decide what kind of results they want.

server/monitoring/sqlite.go
```diff
@@ -90,23 +89,35 @@ func (p *SqliteProvider) GetMetricsListByClientID(ctx context.Context, clientID
 	return val, err
 }
 
-func (p *SqliteProvider) GetMetricsSinceLimitedByClientID(ctx context.Context, clientID string, filters []query.FilterOption) ([]monitoring2.ClientMetricsPayload, error) {
-	q := "SELECT * FROM `measurements` as `metrics` WHERE `client_id` = ? "
+func (p *SqliteProvider) GetMetricsListDownsampledByClientID(ctx context.Context, clientID string, hours float64, o *query.ListOptions) ([]monitoring2.ClientMetricsPayload, error) {
+	q := `SELECT
+		timestamp,
+		avg(cpu_usage_percent) as cpu_usage_percent,
```
domenipavec, line 95: Any reason to not round this the same as memory and io usage averages?

share/types/types.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+// JSONString is a string containing JSON encoded data, it prevents further json encoding when used inside a struct that gets encoded
+type JSONString string
```
domenipavec, line 4: Does `json.RawMessage` not work? If not specify why in the comment.

server/monitoring/service.go
devgek: This format was demanded in the user story, so i sticked to it.

share/query/convert.go
devgek: I will use your pagination-implementation. At the time, when i started to implement the feature, nothing similar existed. So i implemented it also.

db/migration/monitoring/bindata.go
devgek: I thought, that it would be ok to change table design without a new migration script, as long as we are on a  pre-release.
When a feature is officially released, then it's clear to create a migration script for every change.
Am i wrong with this assumption?

server/api.go
devgek: "metrics" is not a list endpoint per se. Default response (without any filters) is the latest measurement, a single result.
With given filters it is a an endpoint, that delivers a list of measurements.

server/monitoring/service.go
domenipavec: I would question Thorsten if there's any reason not to use a standard format.

db/migration/monitoring/bindata.go
domenipavec: Well this could be problematic for everyone that has already run the migrations - so probably devs and test servers.
Not sure if we have a specific policy here, you can ask Thorsten.
In general I would consider everything that is merged to master as deployed and thus requiring a new migration for changes.

server/api.go
domenipavec: I guess if this are the specific requirements then it's ok. But I would question the requirements here, as this behavior will be confusing to the api users - list endpoints with pagination are standard, while returning a single result without filter is not.

devgek: I talked to Thorsten and her agreed with you. I will change the "metrics" endpoint to allways deliver a list, in case of default a list with one entry.

db/migration/monitoring/bindata.go
devgek: Thorsten meant, that it is ok, that we only make migration scripts for changes between official releases. So in this case the last official release was 0.4.2 and measurements table is available in master branch since 0.4.5

server/monitoring/service.go
devgek: Spoke to Thorsten. time.RFC3339 is good, so i will change this ;-)

server/monitoring/service.go
devgek: Yes, that was my intention and i agree with you, it's weird. To make it clear i will change the api to: 
filter[timestamp][gt]=636009200 or otherwise
filter[timestamp][since]=2021-01-01T00:00:00

server/monitoring/service.go
devgek: Thorsten meant, that the decision "downsampling or not" should be made on the server side as it is implemented yet.

share/types/types.go
devgek: Because scan database value to struct fails with: sql: Scan error on column index 1, name "processes": unsupported Scan, storing driver.Value type string into type *json.RawMessage

server/monitoring/sqlite.go
devgek: No reason, just a bug ;-)

server/monitoring/service.go
devgek: Discussed this point once again with Thorsten and convinced him to make separate endpoints for "metrics" and "graph-metrics" (downsampling).
I changed the implementation of "metrics", "mountpoints" and "processes" so that they are very similar to "auditlog".

server/api.go
```diff
@@ -3259,82 +3246,60 @@ func (al *APIListener) handleGetClientMetricsLatest(w http.ResponseWriter, req *
 		al.jsonError(w, err)
 		return
 	}
-	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(clientMetrics))
+	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(payload))
```
domenipavec, line 3249: You shouldn't surround the payload with additional `api.SuccessPayload` when you already have one. The result should be `{"data": [], "meta": {}}`, not `{"data":{"data": [], "meta": {}}}`.

server/api.go
```diff
@@ -3259,82 +3246,60 @@ func (al *APIListener) handleGetClientMetricsLatest(w http.ResponseWriter, req *
...
 			al.jsonErrorResponseWithTitle(w, http.StatusNotFound, fmt.Sprintf("processes for client with id %q not found", clientID))
 			return
 		}
-		al.jsonError(w, serviceErr)
+		al.jsonError(w, err)
 		return
 	}
-
-	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(clientProcesses))
+	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(payload))
```
domenipavec, line 3284: Same

server/api.go
```diff
@@ -3259,82 +3246,60 @@ func (al *APIListener) handleGetClientMetricsLatest(w http.ResponseWriter, req *
...
 			al.jsonErrorResponseWithTitle(w, http.StatusNotFound, fmt.Sprintf("mountpoints for client with id %q not found", clientID))
 			return
 		}
-		al.jsonError(w, serviceErr)
+		al.jsonError(w, err)
 		return
 	}
-
-	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(clientMountpoints))
+	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(payload))
```
domenipavec, line 3302: Same

server/api/monitoring/payload.go
```diff
@@ -1,26 +1,55 @@
 package monitoring
```
domenipavec, line 1: Nit.: I would just put everything in one `monitoring` package instead of having 2 (one in server one in api).

server/monitoring/sqlite.go
```diff
@@ -46,65 +43,86 @@ func NewSqliteProvider(dbPath string, logger *chshare.Logger) (DBProvider, error
...
+		min(cpu_usage_percent) as cpu_usage_percent_min,
+		max(cpu_usage_percent) as cpu_usage_percent_max,
+		round(avg(memory_usage_percent),2) as memory_usage_percent_avg,
+		min(memory_usage_percent) as memory_usage_percent_min,
+		max(memory_usage_percent) as memory_usage_percent_max,
+		round(avg(io_usage_percent),2) as io_usage_percent_avg,
+		min(io_usage_percent) as io_usage_percent_min,
+		max(io_usage_percent) as io_usage_percent_max
+	FROM measurements
+	WHERE client_id = ? and timestamp >= ? and timestamp <= ?
```
domenipavec, line 111: Nit.: There's inconsistency in how `gt` and `since` behave by default if used in query and how they're used in this endpoint (e.g. in used with query package `gt` is using `>` and `since` is using `>=`, while here they're both using `>=`.

domenipavec (commented):
This is mostly ok now, just remove the double `api.SuccessPayload`. Feel free to ignore the nits.

domenipavec (approved):

PR #356, opened 2021-11-19
DEV-2375 tunnel proxy with ssl offloading

devgek:


server/clients/tunnel_proxy.go
```diff
@@ -0,0 +1,134 @@
...
+func NewTunnelProxyRequiredConfig(certFile string, keyFile string) *TunnelProxyConfig {
+	return &TunnelProxyConfig{
+		ProxyRequired: true,
+		cert:          certFile,
+		key:           keyFile,
+	}
+}
+
+func (tpc *TunnelProxyConfig) Addr() string {
+	return tpc.Host + ":" + tpc.Port
```
domenipavec, line 37: You could use `net.JoinHostPort`

server/client_service.go
```diff
@@ -320,6 +329,27 @@ func (s *ClientService) startClientTunnels(client *clients.Client, remotes []*ch
...
+		return clients.NewTunnelProxyNotRequiredConfig(), nil
+	}
+
+	// tunnel proxy required
+	config := clients.NewTunnelProxyRequiredConfig(s.serverConfig.TunnelProxyCertFile, s.serverConfig.TunnelProxyKeyFile)
+	if tunnelACL != nil {
+		proxyACL := *tunnelACL
+		config.ProxyACL = &proxyACL
+	}
+	port, err := s.portDistributor.GetRandomPort()
```
domenipavec, line 343: Shouldn't proxy listen on the `host:port` that user specified for the tunnel, and the actual tunnel listen on `localhost:randomport` where the proxy could access it but not the user?

server/client_service.go
```diff
@@ -320,6 +329,27 @@ func (s *ClientService) startClientTunnels(client *clients.Client, remotes []*ch
...
+	if err != nil {
+		return nil, err
+	}
+	config.Port = strconv.Itoa(port)
```
domenipavec, line 347: This setting of some arguments via `NewTunnelProxyRequiredConfig` constructor and some here directly is a bit weird. Actually this whole method is a bit weird. Why is there not a single constructor for tunnel proxy config that would take remote and acl and construct the config?

server/client_service.go
```diff
@@ -90,10 +91,12 @@ var clientsListDefaultFields = map[string][]string{
 
 // NewClientService returns a new instance of client service.
 func NewClientService(
+	serverConfig *ServerConfig,
```
domenipavec, line 94: Nit.: clients service depending on the whole `ServerConfig` is coupling things quiet a bit. It would be nicer to have a separate struct with just the needed config values. See how I implemented it for `auditlog`.

server/clients/client.go
```diff
@@ -91,19 +92,43 @@ func (c *Client) FindTunnelByRemote(r *chshare.Remote) *Tunnel {
...
 	if t != nil {
 		return t, nil
 	}
 
+	// add localhost to tunnel acl to allow access from tunnel proxy
+	if tunnelProxyConfig.ProxyRequired {
+		if acl == nil {
+			acl, _ = ParseTunnelACL(LocalHost)
+		} else {
+			acl.AddACL(LocalHost)
```
domenipavec, line 106: Wouldn't it be better to completely replace the tunnel ACL with `localhost`, since users shouldn't access it directly, but only through proxy?

server/clients/tunnel_proxy.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+	forward:
+		w.Header().Set("X-RPort-Tunnel-Proxy", tp.config.Addr())
+		p.ServeHTTP(w, r)
+	}
+}
+
+func (tp *TunnelProxy) Start(ctx context.Context) error {
+	forwardTo := *tp.Tunnel.Remote.Scheme + "://" + tp.Tunnel.Remote.LocalHost + ":" + tp.Tunnel.Remote.LocalPort
```
domenipavec, line 82: You should probalby use `url` package for constructing this instead of hardcoding the separators, especially if you're parsing it to url structure anyway.

server/clients/tunnel_proxy.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+
+	tp.Logger.Debugf("create https reverse proxy with ssl offloading forwarding to %s", forwardURL)
+	proxy := httputil.NewSingleHostReverseProxy(forwardURL)
+	sslOffloadingTransport := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true, //nolint:gosec
+			MinVersion:         tls.VersionTLS12,
```
domenipavec, line 93: Why set the min version? If we're not verifying certificates anyway, we're obviously not that concerned with security. So this would only break compatibility with older systems.

server/clients/tunnel_proxy.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}()
+
+	tp.Logger.Debugf("tunnel proxy started")
+	return nil
+}
+
+func (tp *TunnelProxy) Stop(ctx context.Context) error {
+	ctxShutDown, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer func() {
```
domenipavec, line 125: Why not just `defer cancel()` directly?

share/remote.go
```diff
@@ -33,6 +33,8 @@ type Remote struct {
 	Scheme             *string `json:"scheme"`
 	ACL                *string `json:"acl"` // string representation of Tunnel.TunnelACL field
 	IdleTimeoutMinutes int     `json:"idle_timeout_minutes"`
+	HTTPProxy          bool    `json:"http_proxy"`
+	ProxyPort          string  `json:"pport,omitempty"`
```
domenipavec, line 37: I'm not sure this is needed, user's should just see the port where proxy is listening. The tunnel port that proxy connect to is only an implementation detail.

server/clients/tunnel_proxy.go
```diff
@@ -0,0 +1,134 @@
...
+	proxy := httputil.NewSingleHostReverseProxy(forwardURL)
+	sslOffloadingTransport := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true, //nolint:gosec
+			MinVersion:         tls.VersionTLS12,
+		},
+	}
+	proxy.Transport = sslOffloadingTransport
+	proxy.ErrorHandler = func(writer http.ResponseWriter, request *http.Request, err error) {
+		tp.Logger.Debugf("Error during proxy request %v", err)
```
domenipavec, line 98: I'm not sure how many errors are happening, so this is ok if there are a lot. But otherwise I would think errors should be logged at error or at least info level.

server/api.go
```diff
@@ -1065,6 +1065,17 @@ func (al *APIListener) handlePutClientTunnel(w http.ResponseWriter, req *http.Re
 		}
 	}
 
+	httpProxy := req.URL.Query().Get("http_proxy")
+	if httpProxy == "1" && !al.config.Server.tunnelProxyAllowed {
```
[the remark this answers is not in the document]
devgek: done

server/clients/tunnel_proxy.go
```diff
@@ -0,0 +1,134 @@
...
+func (tp *TunnelProxy) tunnelProxyHandlerFunc(p *httputil.ReverseProxy) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if tp.config.ProxyACL != nil {
+			clientIP, _, err := net.SplitHostPort(r.RemoteAddr)
+			if err == nil {
+				ipv4 := net.ParseIP(clientIP)
+				if ipv4 != nil {
+					tcpIP := &net.TCPAddr{IP: ipv4}
+					if tp.config.ProxyACL.CheckAccess(tcpIP) {
+						goto forward
```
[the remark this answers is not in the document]
devgek: done

server/clients/tunnel_proxy.go
devgek: done

server/clients/tunnel_proxy.go
devgek: done

server/clients/tunnel_proxy.go
devgek: done

server/clients/tunnel_proxy.go
devgek: done

server/client_service.go
devgek: done

server/client_service.go
devgek: done

server/client_service.go
devgek: i agree with you, done.

server/clients/client.go
devgek: done

share/remote.go
devgek: done

server/clients/tunnel_proxy.go
devgek: done

domenipavec (approved):

PR #357, opened 2021-11-23
DEV-2237: Store API sessions

m-terel:


domenipavec (dismissed):
The code itself looks good.
But it's a bit weird that we're storing jwt tokens in the database, when the point of signed jwt tokens is that you don't have to store them.

m-terel:
@domenipavec yeah, but there is no other way to keep API sessions alive after server restart :)

PR #358, opened 2021-11-24
DEV-2341 handle server capabilities

devgek:
As a result of the new monitoring feature, the server now sends its capabilities to a client after establishing the connection. On the client side you can do things dependent of server capabilities.


share/capabilities.go
```diff
@@ -0,0 +1,13 @@
...
+
+type Capabilities struct {
+	ServerVersion string
+	Monitoring    int
+}
+
+func NewCapabilities() *Capabilities {
+	return &Capabilities{
+		ServerVersion: BuildVersion,
+		Monitoring:    1,
```
domenipavec, line 11: Not really sure why `monitoring` is an int and not a bool, is it supposed to be a version that can be increased in the future? I would either change it to bool or rename it to `MonitoringVersion` and define `1` as a constant.

client/client.go
```diff
@@ -337,6 +337,26 @@ func (c *Client) sendConnectionRequest(ctx context.Context, sshConn ssh.Conn) er
 	return nil
 }
 
+//afterPutCapabilities is the place to do things dependent on server capabilities
+func (c *Client) afterPutCapabilities(ctx context.Context) {
+	if c.serverCapabilities.Monitoring > 0 {
+		c.monitor.Start(ctx)
```
domenipavec, line 343: What happens after clients loses connection and reconnects? As far as I can see `Start` will make another go routine and you will be sending metrics twice. Either make `Start` make the go routine only the first time (you can use `sync.Once`), or stop the monitoring on disconnect (might be better since the fallback server might not support monitoring?).

share/capabilities.go
```diff
@@ -0,0 +1,13 @@
+package chshare
+
+type Capabilities struct {
```
domenipavec, line 3: Nit.: This might belong in `chshare/comm` or `chshare/models`. Not sure honestly :)

share/capabilities.go
devgek: Your guess was correct, it is a version. Change done.

share/capabilities.go
devgek: Moved to models.

client/client.go
devgek: Monitor is now stopping on client disconnect.

domenipavec (approved):

PR #361, opened 2021-11-26
DEV-2377: Add stored tunnels api

domenipavec:


server/clients/storedtunnels/stored_tunnels.go
```diff
@@ -0,0 +1,106 @@
...
+	provider Provider
+}
+
+func New(db *sqlx.DB) *Manager {
+	return &Manager{
+		provider: newSQLiteProvider(db),
+	}
+}
+
+func (m *Manager) List(r *http.Request, clientID string) (*api.SuccessPayload, error) {
```
devgek, line 49: Only a small note: I would not give the http.Request to Manager.List because than they are coupled. Keeping it independent prepares Manager to be used elsewhere, eg. in client_listener.

devgek (approved):

server/clients/storedtunnels/stored_tunnels.go
domenipavec: Good point, and it's easy to pass `query.ListOptions` instead of request.

devgek (approved):
lgtm

PR #362, opened 2021-11-27
Dev 2358 send config to server

devgek:


server/api.go
```diff
@@ -796,6 +796,7 @@ type ClientPayload struct {
 	AllowedUserGroups      *[]string                `json:"allowed_user_groups,omitempty"`
 	Tunnels                *[]*clients.Tunnel       `json:"tunnels,omitempty"`
 	UpdatesStatus          **models.UpdatesStatus   `json:"updates_status,omitempty"`
+	ClientConfiguration    *chshare.Config          `json:"client_configuration,omitempty"`
```
domenipavec, line 799: If you want to ever return a `null` value for this, then you need to use double pointer as above for updates status. Otherwise if config is `null` and the field is requested, the `omitempty` will cause the response to not contain the field at all.

share/client_config.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"net/http"
+	"net/url"
+	"regexp"
+	"time"
+)
+
+type Config struct {
+	Client         ClientConfig     `mapstructure:"client"`
```
domenipavec, line 11: You should probably set `json` tags for all these, so the json won't have camel case and will match the rest of the api.

client/config.go
```diff
@@ -12,81 +12,22 @@ import (
...
-type CommandsConfig struct {
-	Enabled       bool      `mapstructure:"enabled"`
-	SendBackLimit int       `mapstructure:"send_back_limit"`
-	Allow         []string  `mapstructure:"allow"`
-	Deny          []string  `mapstructure:"deny"`
-	Order         [2]string `mapstructure:"order"`
-
-	allowRegexp []*regexp.Regexp
-	denyRegexp  []*regexp.Regexp
+type ClientConfigHolder struct {
```
domenipavec, line 26: If the only point of `ClientConfigHolder` is to attach parse and validation methods to it, then maybe it would be nicer to just have functions that take `Config` as argument?

share/client_config.go
```diff
@@ -0,0 +1,80 @@
+package chshare
```
domenipavec, line 1: I'm not sure having client's config in share is the best option (especially as a struct named `Config`).
I thought about treating client's config on server as json blob so we don't need this in server, but that's not ideal - if we have it on the server we might want to use it at some point.
Maybe put this in a separate package like `share/clientconfig/client_config.go`? Or at least rename the top level struct to `ClientConfig`.
I would also consider moving server config to the similar place, because having consistency for client and server configs would be nice.

client/config.go
devgek: The reason, why i created ClientConfigHolder were a lot of import cycle problems, after moving Config to package chshare.
So i decided to separate client config data from client config functions. Now chshare.Config is pure data and can be used as data transfer object in different environments (server and client).

share/client_config.go
devgek: I have also thought about that, but had the fear, that this would be too much refactoring for now. Config contains ClientConfig struct in the same package, which had to be refactored also.
I think chshare is the package, where all things are located, that are shared between client and server. As long as we do not need server config on the client side, i would let server config where it is now.

server/api.go
devgek: Done.

share/client_config.go
devgek: Done.

share/client_config.go
devgek: refactored client_config.go to /share/clientconfig/client_config.go

client/config.go
domenipavec: I understand that. I just think that the struct itself is unnecessary. You could have `ParseAndValidate(*Config)` function that takes config as argument instead of method on this struct.

domenipavec (dismissed):

client/monitoring/docker/docker.go
```diff
@@ -5,7 +5,7 @@ import (
 	"runtime"
 	"time"
 
-	chshare "github.com/cloudradar-monitoring/rport/share"
+	chshare "github.com/cloudradar-monitoring/rport/share/logger"
```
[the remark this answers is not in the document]
devgek: Done.

domenipavec (approved):

PR #363, opened 2021-11-30
DEV-2341-set-measurement-timestamp-on-server

devgek:


server/monitoring/service_test.go
```diff
@@ -7,9 +7,29 @@ import (
...
+	defer dbProvider.Close()
+
+	service := NewService(dbProvider)
+	mClient := time.Now().UTC()
+	m := &models.Measurement{
+		ClientID:  "test1",
+		Timestamp: mClient,
+	}
+	sleep := time.Duration(1) * time.Second
+	time.Sleep(sleep)
```
domenipavec, line 26: You should avoid sleep if possible in tests, since it just slows them down unnecessarily. If it's unavoidable then at least make it in millisecond range. Here you could just initialize `mClient` timestamp to `time.Now().Add(-time.Second)` instead of sleeping.

devgek: Done.

domenipavec (approved):

PR #366, opened 2021-12-01
DEV-2397: Support none-English windows

m-terel:


client/encoding/encoding_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/text/encoding/charmap"
+
+	chshare "github.com/cloudradar-monitoring/rport/share"
+)
+
+func TestToUTF8(t *testing.T) {
```
domenipavec, line 14: Would be nice to see test cases with utf-8/ascii text and ones with low confidence.

domenipavec (dismissed):

m-terel:
reimplemented in #373 with a different approach, current wasn't working well

PR #373, opened 2021-12-05
DEV-2397: Support none-English windows

m-terel:
- convert cmd output to utf-8 using windows chcp command


client/system/encoding_win.go
```diff
@@ -0,0 +1,69 @@
...
+	codePageToIANAMapping = map[string]string{
+		"65000": "utf-7",
+		"65001": "utf-8",
+		"1252":  "windows-1252",
+	}
+)
+
+// DetectConsoleEncoding returns encoding that system console is using. Returns nil, if it's UTF-8
+func DetectConsoleEncoding(ctx context.Context) (encoding.Encoding, error) {
+	cmd := exec.CommandContext(ctx, "cmd", "/c", "chcp")
```
domenipavec, line 28: I'm pretty sure that `Run` should be called on `Cmd` for it to be executed, right?

client/system/encoding_win_test.go
```diff
@@ -0,0 +1,75 @@
+//go:build windows
+// +build windows
```
domenipavec, line 2: Since we only run tests on ubuntu in github actions and the code you're testing doesn't actually rely on anything that couldn't run on linux, I think it would be better to have this code compiled for all OSs.

client/system/encoding_win.go
m-terel: Nope :)
```
// Output runs the command and returns its standard output.
// Any returned error will usually be of type *ExitError.
// If c.Stderr was nil, Output populates ExitError.Stderr.
```

client/system/encoding_win_test.go
m-terel: moved `detectEncodingByCHCPOutput` to `encoding.go`

client/system/encoding_win.go
domenipavec: Interesting, I've always used `Run` :)

domenipavec (approved):

PR #394, opened 2021-12-22
Interpreter abs path and interpreter aliases

breathbath:


.github/workflows/go.yml
```diff
@@ -16,10 +16,18 @@ jobs:
...
       - name: Test
         run: go test -race -v ./...
+
+  test-win:
+    name: Test Client Windows
+    runs-on: windows-2019
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test
+        run: go test -race -v ./client/...
```
domenipavec, line 33: Any reason not to just run the whole test suite here?

client/system/exec_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"os"
+	"testing"
+
+	"github.com/cloudradar-monitoring/rport/share/logger"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var testLog = logger.NewLogger("client-system", logger.LogOutput{File: os.Stdout}, logger.LogLevelDebug)
```
domenipavec, line 13: There's no reason to have this as a global variable.

client/system/exec_nix_test.go
```diff
@@ -0,0 +1,58 @@
...
+func getCmdBuildTestcases() []cmdBuildTestCase {
+	return []cmdBuildTestCase{
+		{
+			name:        "empty",
+			interpreter: chshare.UnixShell,
+			wantCmdStr:  "/bin/sh -c /script.sh",
+			command:     "/script.sh",
+		},
+		{
+			name:        "non empty sh",
```
domenipavec, line 18: This seems the same as previous?

domenipavec (dismissed):

.github/workflows/go.yml
breathbath: yeah it's a unittests execution for Windows which is actual for client rather than for the server

client/system/exec_test.go
breathbath: no reason - will refactor

client/system/exec_nix_test.go
breathbath: fixed

domenipavec (approved):

domenipavec (approved):

PR #397, opened 2021-12-23
Dev 2384 monitoring net bandwidth

devgek:
Adding network bandwidth to monitoring data and expand the graph-metrics api


client/monitoring/monitor.go
```diff
@@ -138,6 +148,7 @@ func (m *Monitor) sendMeasurement() {
 		}
 
 		m.logger.Debugf("sending %d bytes of monitoring data to server", len(data))
+		m.logger.Debugf("%s", string(data))
```
domenipavec, line 151: You probably don't want to log the whole data.

client/monitoring/networking/cagent_wrapper_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewCAgentNetWatcher(t *testing.T) {
+	netWatcher := NewCAgentNetWatcher()
+	assert.NotNil(t, netWatcher)
+	excludeRegexSlice := netWatcher.InterfaceExcludeRegexCompiled()
+	assert.NotNil(t, excludeRegexSlice)
+	assert.Equal(t, 5, len(excludeRegexSlice))
```
domenipavec, line 14: I think this test is too implementation dependent. There's nothing about 5 regex values that I would consider correct, so if we change the configuration we shouldn't have to change this test as well.

client/monitoring/networking/net_handler.go
```diff
@@ -0,0 +1,59 @@
...
+
+	var mm map[string]interface{}
+	mm, err := nh.netWatcher.Results()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var netLan *models.NetBytes
+	if nh.config.LanCard != nil {
+		netLan = getNetBytesFromMap(mm, inBytesName+"."+nh.config.LanCard.Name, outBytesName+"."+nh.config.LanCard.Name)
```
domenipavec, line 34: Nit.: Instead of doing the same string concat here and below, maybe you should do it in the function and just pass interface name as argument.

client/monitoring/networking/net_handler_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/cloudradar-monitoring/rport/share/clientconfig"
+	"github.com/cloudradar-monitoring/rport/share/models"
+)
+
+func TestNewNetHandler(t *testing.T) {
```
domenipavec, line 12: Instead of doing these trivial tests where you assert something is nil, you should test the main logic. For `net_handler.go` I think it is getting the correct values from the map returned by `netWatcher`.

cmd/rport/main.go
```diff
@@ -221,6 +224,8 @@ func init() {
 	pFlags.Bool("monitoring-pm-enabled", false, "")
 	pFlags.Bool("monitoring-pm-kerneltasks-enabled", false, "")
 	pFlags.Int("monitoring-pm-max-number-processes", 0, "")
+	pFlags.StringArray("monitoring-net-lan", []string{}, "")
```
domenipavec, line 227: I'm a bit confused about this flags. At first I thought that slice was for specifying multiple interfaces. But now I see one needs to specify name and max speed. What is max speed?

cmd/rport/main.go
```diff
@@ -156,6 +156,9 @@ var clientHelp = `
    --monitoring-pm-kerneltasks-enabled, enable or disable monitoring of kerneltasks
    --monitoring-pm-max-number-processes, maximum number of processes in process monitoring list
 
+   --monitoring-net-lan, enable monitoring of lan network card
+   --monitoring-net-wan, enable monitoring of wan network card
```
domenipavec, line 160: Why is the naming `lan` and `wan`? Usually computers don't have wan and lan cards unless they're functioning as routers.
Could we just have a list of interfaces instead of 2 specific?

db/migration/monitoring/sql/002_add_net.down.sql
```diff
@@ -0,0 +1,7 @@
+-- ----------------------------
```
domenipavec, line 1: You seem to have some duplicated files here `002` and `003`.

db/migration/monitoring/sql/002_add_net.down.sql
```diff
@@ -0,0 +1,7 @@
...
+-- drop net columns
+-- ----------------------------
+ALTER TABLE "measurements" DROP COLUMN "net_lan_in";
```
domenipavec, line 4: `sqlite` doesn't support `drop column` as far as I know.

share/query/request_info.go
```diff
@@ -0,0 +1,20 @@
+package query
+
+import "net/http"
+
+type RequestInfo struct {
+	URL string
+}
+
+func ParseRequestInfo(req *http.Request) *RequestInfo {
```
domenipavec, line 9: This could easily be wrong if rport is behind a reverse proxy. I think it would be better to just return relative paths if we need to do that at all.

server/monitoring/sqlite.go
```diff
@@ -124,13 +126,60 @@ func (p *SqliteProvider) ListGraphMetricsByClientID(ctx context.Context, clientI
...
+		min(` + field + `) as ` + alias + `_min,
+		max(` + field + `) as ` + alias + `_max`
+	}
+	q = q + ` 
+	FROM measurements WHERE client_id = ?`
+
+	q, params = query.AddWhere(lo.Filters, q, params)
+
+	q = q + ` GROUP BY round((strftime('%s',timestamp)/(?)),0)`
+	divisor := (math.Round(hours*100) / 100) * 29
```
domenipavec, line 158: Can you add a comment what this does and why it's multiplied by 29?

server/monitoring/service.go
```diff
@@ -52,7 +54,136 @@ func (s *monitoringService) DeleteMeasurementsOlderThan(ctx context.Context, per
...
+	if strings.HasPrefix(graph, "net_usage_percent_") {
+		calculatePercentValues(&entries, bytesMaxLan, bytesMaxWan)
+	}
+
+	return &api.SuccessPayload{
+		Data: entries,
+	}, nil
+}
+
+func calculatePercentValues(entries *[]*ClientGraphMetricsGraphPayload, bytesMaxLan float64, bytesMaxWan float64) {
```
domenipavec, line 107: It's a bit weird to return bandwidth values as percentages. I would expect byte values. Is this a product requirement?

server/monitoring/service.go
```diff
@@ -52,7 +54,136 @@ func (s *monitoringService) DeleteMeasurementsOlderThan(ctx context.Context, per
...
+	if err != nil {
+		return nil, err
+	}
+
+	entries, err := s.DBProvider.ListGraphMetricsByClientID(ctx, clientID, span.Hours(), lo)
+	if err != nil {
+		return nil, err
+	}
+
+	links := &GraphMetricsLinksPayload{
```
domenipavec, line 68: Are these links really useful? Seems there's only a couple of possible options for graph name that can be seen in api documentation.

server/api.go
```diff
@@ -3555,9 +3557,66 @@ func (al *APIListener) handleGetClientGraphMetrics(w http.ResponseWriter, req *h
...
+	}
+	netLan := client.ClientConfiguration.Monitoring.LanCard != nil
+	netWan := client.ClientConfiguration.Monitoring.WanCard != nil
+	if strings.HasSuffix(graph, "_lan") && !netLan ||
+		strings.HasSuffix(graph, "_wan") && !netWan {
+		al.jsonErrorResponseWithTitle(w, http.StatusNotFound, fmt.Sprintf("graph data %s not available for client with id %s", graph, clientID))
+		return
+	}
+
+	bytesMaxLan := 125000.0
```
domenipavec, line 3611: This should be a constant somewhere. Also would be nice to have a comment explaining why this value is `125000`, it's not immediately clear to me.

server/api.go
```diff
@@ -3555,9 +3557,66 @@ func (al *APIListener) handleGetClientGraphMetrics(w http.ResponseWriter, req *h
...
+	if netLan {
+		bytesMaxLan = bytesMaxLan * float64(client.ClientConfiguration.Monitoring.LanCard.MaxSpeed)
```
domenipavec, line 3613: Why are these values computed here? I would expect `monitoringService` to have access to config and have that there.

client/monitoring/monitor.go
devgek: done

client/monitoring/networking/cagent_wrapper_test.go
devgek: done

client/monitoring/networking/net_handler.go
devgek: done

cmd/rport/main.go
devgek: It is the maximum bandwitdth of the network interface. This value is needed to calculate percentage values, which are the main requirement of this user story.

server/monitoring/sqlite.go
devgek: This is the part of "downsampling graph data" (group together graph points, so that you don't get too much points in one request. The value of "29" comes from Thorsten. He did some research and found out that "29" would be the best fit.

server/monitoring/service.go
devgek: Yes, it's a requirement. The api delivers both absolute values in bytes as well as percentage values. The usage of network is better described by percentage values than absolute values.

server/monitoring/service.go
devgek: Also a requirement by Thorsten. He wanted this kind of api.

server/api.go
devgek: done, it's used to convert MBits per second to Bytes per second

server/api.go
devgek: done

cmd/rport/main.go
devgek: It was a requirement in the user story

db/migration/monitoring/sql/002_add_net.down.sql
devgek: done

db/migration/monitoring/sql/002_add_net.down.sql
devgek: It is supported since 3.35.0

share/query/request_info.go
devgek: I think the specs in jsonapi.org require absolute URL's.

client/monitoring/networking/net_handler_test.go
devgek: done

db/migration/monitoring/sql/002_add_net.down.sql
domenipavec: Users might not have the latest version, it doesn't work on my ubuntu machine. But I'm ok if you leave it, since it's in down migration.

domenipavec (approved):

PR #414, opened 2022-01-08
DEV-2393: Set tunnels scheme, proxy and host-header via rport client

m-terel:


cmd/rport/main.go
```diff
@@ -159,6 +159,10 @@ var clientHelp = `
    --monitoring-net-lan, enable monitoring of lan network card
    --monitoring-net-wan, enable monitoring of wan network card
 
+    --scheme, An optional scheme value that is used to create the given tunnels(<remote>s). For example, 'http', ssh', 'rdp', etc.
```
domenipavec, line 162: Are we sure that having global settings for these is good enough? Wouldn't  it be better to enable this settings for each tunnel separately? Maybe using url-like syntax: `https://127.0.0.1:8000:127.0.0.1:8000/tcp?reverse-proxy=true&host-header=abc.com`

domenipavec (dismissed):
Code looks good, but I would question the decision to make the settings global instead of per tunnel.

cmd/rport/main.go
m-terel: I've discussed it with Thorsten. He prefers the current behavior.

PR #446, opened 2022-02-12
DEV-2469: Tunnel expiry after given timeout

m-terel:


server/validation/tunnel.go
```diff
@@ -56,3 +55,38 @@ func ResolveIdleTunnelTimeoutValue(idleTimeoutMinutesStr string, skipIdleTimeout
...
+	dur, err := time.ParseDuration(durationStr)
+	if err != nil {
+		return 0, errors2.APIError{
+			Message:    "invalid tunnel auto-close format",
+			Err:        err,
+			HTTPStatus: http.StatusBadRequest,
+		}
+	}
+
+	if dur < 0 {
```
domenipavec, line 77: I think the value equal to `0` should be allowed to explicitly disable auto close.

domenipavec (approved):

server/validation/tunnel.go
m-terel: Actually, it's allowed :) It can be 0 and it means - no auto-close.

PR #451, opened 2022-02-15
DEV-2454 rdp proxy fixes for nla security

devgek:


server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
```diff
@@ -59,6 +72,47 @@ func (tc *TunnelProxyConnectorRDP) serveIndex(w http.ResponseWriter, r *http.Req
...
+
+func parseGuacToken(r *http.Request) (*GuacToken, error) {
+	var query url.Values
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+	_ = r.Body.Close()
+	queryString := string(data)
+	query, err = url.ParseQuery(queryString)
```
domenipavec, line 100: It's extremely unusual to have url query data in request body. Either use json or http form data.

server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
```diff
@@ -71,12 +125,21 @@ func (tc *TunnelProxyConnectorRDP) connectToGuacamole(request *http.Request) (gu
...
-	if username != "" {
-		config.Parameters["username"] = username
+
+	token := query.Get(queryParToken)
+	guacToken := tc.guacTokenStore.Get(token)
+	if guacToken == nil {
+		tc.tunnelProxy.Logger.Errorf("Cannot find guac token %s", token)
+		return nil, err
 	}
+	defer tc.guacTokenStore.Delete(token)
```
domenipavec, line 135: What's the point of using defer? You're not getting the value from the store at any point after this, you might as well delete immediately.

server/clients/clienttunnel/guac_token.go
```diff
@@ -0,0 +1,43 @@
...
+type GuacToken struct {
+	username     string
+	password     string
+	security     string
+	width        string
+	height       string
+	serverLayout string
+}
+
+type GuacTokenStore struct {
```
domenipavec, line 16: It seems a bit unnecessary to be storing this information on the server, especially as it's only stored temporarily. What's the reason for changing away from passing data as query params on connect directly?

server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
devgek: That's what it is, http form data transmitted via a "POST"-Request.

server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
devgek: You'r right. Defer is not necessary. So i will delete immediately.

server/clients/clienttunnel/guac_token.go
devgek: The reason is that the javascript-library (guacamole-common-js) which is used to start the websocket-tunnel to the Guacamole server in "client.connect" initiates a "GET"-Request and not a "POST"-Request. And because we are now sending the login-password as part of this request, the password is sent as clear text as part of the query string. So i decided to do an extra POST-Request prior to the websocket-tunnel - connect-request, which stores all the used parameters (but especially the password) in GuacToken. The following websocket-connect-request, which is the "GET"-Request only sends the uuid of the token as part of the query string.

server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
domenipavec: Then why not use https://pkg.go.dev/net/http#Request.ParseForm

server/clients/clienttunnel/guac_token.go
domenipavec: It would be good to add a comment describing this.

server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
devgek: yes, done. ;-)

server/clients/clienttunnel/guac_token.go
devgek: done

thorstenkramm (approved):

PR #465, opened 2022-03-01
Initial Dockerfile for rportd

nwmcsween:
This uses some interesting scripting in the init script to pass env variables as args to rportd, might need a bit more testing to ensure everything is working correctly.


nwmcsween:
The problem with mutating a config file is handling implicit dependencies on settings, do you regenerate the entire config file or only part of it and will there be a possible matrix of settings that create a insecure config? The arg approach also allows having a read only container with a /var/lib/rport volume mount.

AC-WhiteGlint:
You only expose port 8080 and 3000 but what about the necessary port for the tunnels? (by default 20000 to 30000)

nwmcsween:
> You only expose port 8080 and 3000 but what about the necessary port for the tunnels? (by default 20000 to 30000) also these option should be exposed :
> 
> * 2fa
>   
>   * smtp
>   * topt
> * database
>   -sqllite3
>   -mysql
>   etc.... (basically all the setting available in the conf file)
>   You should also install fail2ban by default  and give the possibility to the user :
> * Enable or disable fail2ban
> * change max retry
> * change bantime

EXPOSE doesn't really do anything in a Dockerfile it's just for documenting that exposed static ports. SMTP, TOPT, *SQL should work just add the correct ENV args. Fail2ban should be external, you can log to journald, etc and block with fail2ban, crowdsec in another container or on the host itself.

thorstenkramm:
@nwmcsween As mentioned, we are not yet ready to support docker. And the comment from @AC-WhiteGlint is justified. There are many things to consider when running the rport server in a docker container. 
Any contribution is welcome, but this pull request not mature enough. Sorry for rejecting. 
Why not move the Dockerfile to a [Gist](https://gist.github.com/), so people can discuss and contribute there!

PR #480, opened 2022-03-21
accept only TLSv1.3 and use strong cipher suites

thorstenkramm:


domenipavec (dismissed):
You should fix the failing test  in `http_server_test.go` (missing argument to `WithTLS`).
Otherwise looks good :+1:

share/http_server_test.go
```diff
@@ -15,7 +16,13 @@ func TestNewHttpServer(t *testing.T) {
 }
 
 func TestNewHttpServerWithTLS(t *testing.T) {
-	s := NewHTTPServer(123, WithTLS("test.crt", "test.key"))
+	tlsConfig := &tls.Config{
```
domenipavec, line 19: You should probably just use `security.TLSConfig` instead of re-declaring it here.

domenipavec (approved):

PR #483, opened 2022-03-22
Just rewrite known vue paths, return 404 otherwise

thorstenkramm:


server/api/middleware/rewrite404_test.go
```diff
@@ -52,3 +52,40 @@ func TestRewrite404(t *testing.T) {
...
+		t.Run(tc.Path, func(t *testing.T) {
+			mockHandler := func(w http.ResponseWriter, r *http.Request) {
+				if r.URL.Path != "/" {
+					http.NotFound(w, r)
+					return
+				}
+				_, err := w.Write([]byte("hello world"))
+				require.NoError(t, err)
+			}
+			vueHistoryPaths := []string{"dashboard"}
```
domenipavec, line 80: I would add 2 entries here with 2nd being `dashboard`, so you're testing that the code checks all entries not just the 1st.

server/api.go
```diff
@@ -297,7 +297,18 @@ func (al *APIListener) initRouter() {
 
 	docRoot := al.config.API.DocRoot
 	if docRoot != "" {
-		r.PathPrefix("/").Handler(middleware.Rewrite404(http.FileServer(http.Dir(docRoot)), "/"))
+		// Start a http file server with proper Vue.js HTML5 history mode (aka rewrite to /) for the following paths
+		vueHistoryPaths := []string{
```
domenipavec, line 301: I would define these as a global variable at the top of the file, so it's easier to find and change in the future.

domenipavec (approved):
Two small comments, otherwise looks good.

domenipavec (approved):

PR #495, opened 2022-04-06
Accept window title as query parameter

thorstenkramm:


server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
```diff
@@ -79,6 +80,7 @@ func (tc *TunnelProxyConnectorRDP) serveIndex(w http.ResponseWriter, r *http.Req
 		queryParKeyboard:  r.Form.Get(queryParKeyboard),
 		queryParWidth:     r.Form.Get(queryParWidth),
 		queryParHeight:    r.Form.Get(queryParHeight),
+		queryParTitle:     r.Form.Get(queryParTitle),
```
domenipavec, line 83: You might want to add a default if this is not set.

domenipavec (approved):

server/clients/clienttunnel/tunnel_proxy_connector_rdp.go
thorstenkramm: On the index.html template, the string " RDP" is always appended, so the title will never be empty.

PR #496, opened 2022-04-07
Introduced new config for the pairing service url

thorstenkramm:
@domenipavec Tests are still missing.


server/config.go
```diff
@@ -486,19 +487,38 @@ func (s *ServerConfig) parseAndValidateURLs() error {
...
+	} else {
+		if err := validateHTTPorHTTPSURL(s.PairingURL, "pairing"); err != nil {
+			return err
 		}
 	}
 
 	return nil
 }
 
+func validateHTTPorHTTPSURL(testURL string, name string) error {
```
domenipavec, line 506: You should use `errors.Wrap` in calling function to add details to the error instead of passing in the name.

server/config_load_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+func TestLoadingExampleConf(t *testing.T) {
+	var (
+		viperCfg *viper.Viper
+		cfg      = &Config{}
+	)
+	viperCfg = viper.New()
+	viperCfg.SetConfigType("toml")
+	viperCfg.SetConfigFile("../rportd.example.conf")
+	path, _ := os.Getwd()
```
domenipavec, line 22: You shouldn't ignore errors, but use `require.NoError` instead.

server/config_test.go
```diff
@@ -483,15 +482,17 @@ func TestParseAndValidateAPI(t *testing.T) {
...
+			ExpectedError: "API: conflicting 2FA configuration, two factor auth and totp_enabled options cannot be both enabled",
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.Name, func(t *testing.T) {
 			tc.Config.Server = defaultValidMinServerConfig
 			err := tc.Config.ParseAndValidate()
-			assert.Equal(t, tc.ExpectedError, err)
+			if len(tc.ExpectedError) != 0 {
```
domenipavec, line 493: For strings it's cleaner to check against empty string: `tc.ExpectedError != ""`

server/config_test.go
```diff
@@ -483,15 +482,17 @@ func TestParseAndValidateAPI(t *testing.T) {
...
+				assert.EqualError(t, err, tc.ExpectedError)
+			}
```
domenipavec, line 495: Add else clause with `assert.NoError`.

server/config_test.go
```diff
@@ -181,7 +218,9 @@ func TestParseAndValidateClientAuth(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.Name, func(t *testing.T) {
 			err := tc.Config.parseAndValidateClientAuth()
-			assert.Equal(t, tc.ExpectedError, err)
+			if err != nil {
```
domenipavec, line 221: You should branch based on `tc.ExpectedError != ""`, you can't be sure what the `err` is.
Also add else clause with `assert.NoError`, so that `err` is always checked.

domenipavec (approved):

PR #534, opened 2022-05-03
Introduced new tunnel_host setting resolves #469

### LINKED ISSUE #469: Feature : new config entry to define the ip/domain where the tunnel is located
Hi,

I came accross an issue when using rport api behin a https reverse proxy who is also behind cloudflare which I am using to protect all my app.
When connecting to the web console and creating a tunnel it used the FQDN, the problem is that the api is behing cloudflare services, which then resolve to cldouflare ips.
It could be greate to have anohter entry in the config to specific the ip/fqdn where the tunnel is so that we can connect to the servers/computer even when using service like cloudflare to protect the api.

thorstenkramm:


server/config.go
```diff
@@ -499,6 +505,13 @@ func (s *ServerConfig) parseAndValidateURLs() error {
 	return nil
 }
 
+func (s *ServerConfig) parseAndValidateHosts() error {
+	if len(s.TunnelHost) != 0 && !govalidator.IsDNSName(s.TunnelHost) && !govalidator.IsIP(s.TunnelHost) {
```
domenipavec, line 509: `s.TunnelHost != ""` is more idiomatic than using `len` in go.

server/config_test.go
```diff
@@ -43,6 +43,19 @@ func TestParseAndValidateServerConfig(t *testing.T) {
...
+				Server: ServerConfig{
+					URL:          []string{"http://localhost/"},
+					DataDir:      "./",
+					Auth:         "abc:def",
+					UsedPortsRaw: []string{"10-20"},
+					TunnelHost:   "bad tunnel host",
+				},
+			},
+			ExpectedError: "invalid tunnel_host 'bad tunnel host': use IP address or FQDN",
+		},
```
domenipavec, line 58: It might be nice to add a non-empty valid host test case, so that both cases are covered.

domenipavec (approved):

domenipavec (approved):

PR #540, opened 2022-05-11
Dev 2527

thorstenkramm:
@domenipavec This is my first approach for a global WAL switch.
I'm thinking about extending DataSourceOptions struct to
```
type DataSourceOptions struct {
	WALEnabled bool
	BaseDir    string
}
```
Then we can remove all the repeated `path.Join(config.Server.DataDir, "<DB_NAME>")` and the db dir is injected just once by `GetSQLiteDataSourceOptions()` in server/config.go. What's your opinion?


db/migration/dummy/sql/001_init.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE dummy;
```
domenipavec, line 1: You're creating table `clients` but dropping table `dummy`.

server/config.go
```diff
@@ -256,6 +259,13 @@ func (c *Config) GetUploadDir() string {
 	return filepath.Join(c.Server.DataDir, files.DefaultUploadTempFolder)
 }
 
+func (s *ServerConfig) GetSQLiteDataSourceOptions() sqlite.DataSourceOptions {
+	if s.SqliteWAL {
+		return sqlite.DataSourceOptions{WALEnabled: true}
+	}
+	return sqlite.DataSourceOptions{}
```
domenipavec, line 266: You could just return `DataSourceOptions{WALEnabled: s.SqliteWAL}` instead of having an if.

domenipavec (approved):
Yes I think adding basedir to options is a good idea.

PR #555, opened 2022-05-31
DEV-2528 implementation of filters for client-auth

thorstenkramm:


server/server.go
```diff
@@ -205,21 +205,11 @@ func NewServer(config *Config, filesAPI files.FileAPI) (*Server, error) {
...
 	}
 
 	if config.Server.AuthFile != "" {
-		fileProvider := clientsauth.NewFileProvider(config.Server.AuthFile)
-		cachedProvider, err := clientsauth.NewCachedProvider(fileProvider)
-		if err != nil {
-			return nil, err
-		}
-		return cachedProvider, nil
+		return clientsauth.NewFileProvider(config.Server.AuthFile), nil
```
domenipavec, line 212: I think that for `FileProvider` we should still use cache, otherwise the whole json file must be loaded on every client connection, which could be potentially expensive.

server/api_clientauth_test.go
```diff
@@ -0,0 +1,543 @@
...
+			provider:        clientsauth.NewDatabaseMockProvider([]*clientsauth.ClientAuth{cl1, cl2, cl3}),
+			wantStatusCode:  http.StatusOK,
+			wantClientsAuth: []*clientsauth.ClientAuth{},
+			wantCount:       0,
+			idFilter:        "sdfetzuj",
+		},
+	}
+
+	for _, tc := range testCases {
+		msg := fmt.Sprintf("test case: %q", tc.descr)
```
domenipavec, line 97: Please use `t.Run` to add `descr` to test name and remove `msg`.

server/api_clientauth_test.go
```diff
@@ -0,0 +1,543 @@
...
+
+		provider clientsauth.Provider
+
+		wantStatusCode  int
+		wantClientsAuth []*clientsauth.ClientAuth
+		wantErrCode     string
+		wantErrTitle    string
+		wantCount       int
+		idFilter        string
+	}{
```
domenipavec, line 45: It would be good to add test cases for empty id filter, successful filter with one result.

server/clientsauth/clientauth.go
```diff
@@ -1,19 +1,25 @@
 package clientsauth
 
-import "sort"
+var SupportedFilters = map[string]bool{
+	"id": true,
+}
+
+var SupportedSorts = map[string]bool{}
+
+var SupportedFields = map[string]map[string]bool{
```
domenipavec, line 9: You can pass `nil` for supported and default fields to remove support for fields on this endpoint, since it makes no sense with only `id`.

server/api_clientauth_test.go
```diff
@@ -0,0 +1,543 @@
...
+			requestBody:     composeRequestBody(cl4.ID, cl4.Password),
+			wantStatusCode:  http.StatusMethodNotAllowed,
+			wantErrCode:     ErrCodeClientAuthSingleClient,
+			wantErrTitle:    "Client authentication is enabled only for a single user.",
+			wantClientsAuth: []*clientsauth.ClientAuth{cl1},
+		},
+	}
+
+	for _, tc := range testCases {
+		msg := fmt.Sprintf("test case: %q", tc.descr)
```
domenipavec, line 309: Same as above.

domenipavec (changes requested):

server/clientsauth/database.go
```diff
@@ -26,15 +32,56 @@ func NewDatabaseProvider(DB *sqlx.DB, tableName string) *DatabaseProvider {
 	}
 }
 
+func NewDatabaseMockProvider(clients []*ClientAuth) *DatabaseProvider {
```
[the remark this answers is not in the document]
thorstenkramm: @domenipavec As far as I know, a function inside a _test file, has a file-only scope. One test cannot use functions declared in another test. Moving the mock function `NewDatabaseMockProvider` into a separated _test file seems not possible. Or am I missing something?

server/api_clientauth_test.go
```diff
@@ -94,54 +111,55 @@ func TestHandleGetClientsAuth(t *testing.T) {
...
+			q.Add("page[limit]", "3")
+			q.Add("filter[id]", tc.idFilter)
+			req.URL.RawQuery = q.Encode()
+			t.Logf("%s: URL tested: %s", tc.descr, req.URL.String())
+			handler := http.HandlerFunc(al.handleGetClientsAuth)
+			w := httptest.NewRecorder()
+			handler.ServeHTTP(w, req)
+
+			// then
+			msg := fmt.Sprintf("test case: %q", tc.descr)
```
domenipavec, line 138: Msg is not required anymore since you're using t.Run

server/api_clientauth_test.go
```diff
@@ -306,59 +324,55 @@ func TestHandlePostClientsAuth(t *testing.T) {
...
+			req := httptest.NewRequest(http.MethodPost, "/api/v1/clients-auth", tc.requestBody)
+
+			// when
+			handler := http.HandlerFunc(al.handlePostClientsAuth)
+			w := httptest.NewRecorder()
+			handler.ServeHTTP(w, req)
+			t.Logf("Got response %s", w.Body)
+
+			// then
+			msg := fmt.Sprintf("test case: %q", tc.descr)
```
domenipavec, line 351: Same

server/clientsauth/file.go
```diff
@@ -5,52 +5,34 @@ import (
...
-	for _, v := range clients {
-		clientAuth[v.ID] = v.Password
-	}
-	cj, _ := json.Marshal(clientAuth)
-	if _, err := f.Write(cj); err != nil {
-		log.Fatalln("error writing client-auth.json for mock provider", err)
-	}
-	return &FileProvider{
-		fileName: authFile,
+		cache:    cache,
```
domenipavec, line 30: You need to either delete or update cache on `Add`

share/query/convert.go
```diff
@@ -37,7 +37,18 @@ func AddWhere(filterOptions []FilterOption, q string, params []interface{}) (str
 		orParts := make([]string, 0, len(filterOptions[i].Values))
 		for _, col := range filterOptions[i].Column {
 			for _, val := range filterOptions[i].Values {
-				part := fmt.Sprintf("%s %s ?", col, filterOptions[i].Operator.Code())
+				var operatorCode = filterOptions[i].Operator.Code()
+				var operatorOption = ""
+				if operatorCode == "=" && strings.Contains(val, "*") {
+					// Implement a SQL LIKE search triggered by a wildcard
+					operatorCode = "LIKE"
```
domenipavec, line 44: This will do case sensitive comparison, while the non db code will do case insensitive. We should do the same here.

Also I implemented the same thing in my PR, I think my implementation is a bit cleaner: https://github.com/cloudradar-monitoring/rport/pull/562/files#diff-d8c93b1d3989c57232dc1cce22fdb2c250a647f6a708a21bdaf6096291ba74e1R43 
If you agree merge my PR and rebase this on top.

thorstenkramm: @domenipavec I adopted your approach. I appended escaping the `%` so users can't use this sign to trigger a wildcard search.

domenipavec (dismissed):

PR #559, opened 2022-06-03
Dev 2541

thorstenkramm:


server/upload.go
```diff
@@ -316,3 +332,29 @@ func validateUploadRequest(ur *UploadRequest) error {
 
 	return ur.Validate()
 }
+
+func validateUserRights(ur *UploadRequest, user *users.User) error {
```
domenipavec, line 336: We shouldn't have multiple functions for checking user access to clients. There already exists 'ClientService.CheckClientsAccess` that does this, you should use it.

api-doc.yml
```diff
@@ -1023,7 +1023,7 @@ paths:
     get:
       tags:
         - "Commands"
-      summary: "Return a short info about all multi-client commands"
+      summary: "Return a short info about all multi-client commands executed by the scheduler"
```
domenipavec, line 1026: I don't think the change is true, it returns all multi client commands not only those by the scheduler.

thorstenkramm: It lands on `func (p *SqliteProvider) GetMultiJobSummaries()` which selects from `multi_jobs`.
The regular script execution not triggered by the schedules stores all data in the table `jobs`. 
I think we should rework this with a clear naming of the endpoints. 
Having `GET /jobs` and `GET /scheduler-jobs` would make it clearer.

domenipavec: That's not true, `multi_jobs` is used any time command/script is executed on multiple clients, that includes both scheduled and manual execution.

domenipavec (dismissed):

PR #578, opened 2022-06-15
Dev 2548

thorstenkramm:


server/client_listener.go
```diff
@@ -216,11 +216,11 @@ func (cl *ClientListener) handleWebsocket(w http.ResponseWriter, req *http.Reque
 	var r *ssh.Request
 	select {
 	case r = <-reqs:
-	case <-time.After(3 * time.Minute):
-		clog.Errorf("SSH connection timeout exceeded")
+	case <-time.After(90 * time.Second):
```
domenipavec, line 219: It would be better to extract `90*time.Second` into a constant and use it here and in the log below, so it's easier to change and keep consistent.

domenipavec (dismissed):

PR #592, opened 2022-06-23
Dev 2570

thorstenkramm:


client/client_test.go
```diff
@@ -408,7 +408,7 @@ func (m *mockServer) WaitForStatus(isConnected bool) error {
 		if m.IsConnected() == isConnected {
 			return nil
 		}
-		time.Sleep(time.Millisecond)
+		time.Sleep(10 * time.Millisecond)
```
domenipavec, line 411: Does this change the results of the tests? It shouldn't.
Having a smaller constant should make the tests run faster, so I don't think we should change this.

domenipavec: You should increase the maximum value for the loop (500) instead of the step.

We can try if that helps, but I don't think that's the problem. I think there might be a race or lock happening somewhere that causes this timeout. I don't have a Windows dev machine to try and debug it.

domenipavec (dismissed):

PR #626, opened 2022-07-11
first draft of server task to  clean up disconnected clients

thorstenkramm:
@domenipavec This is just a draft. Many things are missing. Please have a look at the concept and how the ping jobs are dispatched as go routines. The very verbose logging will be removed later. I didn't want to spend too much time now, if the approach is wrong.


server/client_cleanup.go
```diff
@@ -0,0 +1,101 @@
...
+		t.log.Errorf("Failed to get list of clients.")
+		return err
+	}
+	t.log.Debugf("Starting NewDisconnectedHostsCleanupTask task for %d clients now.", len(clients))
+	//jclients, _ := json.Marshal(clients)
+	//t.log.Debugf("Client List: %s", jclients)
+	var wg sync.WaitGroup
+	for _, client := range clients {
+		t.log.Debugf("Running ping task for client %s %s", client.ID, client.Name)
+		activeClient, err := t.cs.GetActiveByID(client.ID)
```
domenipavec, line 41: There's already `GetAllActive` implemented in `ClientRepository`, you should just expose that on client service instead of checking each client.

server/client_listener.go
```diff
@@ -399,6 +399,7 @@ func (cl *ClientListener) handleSSHRequests(clientLog *logger.Logger, clientID s
 		switch r.Type {
 		case comm.RequestTypePing:
 			_ = r.Reply(true, nil)
```
domenipavec, line 401: Since we're already receiving pings from clients, you could save last ping time here (no need to save in db, just put it on the Client struct) and use that instead of sending pings both ways.

thorstenkramm: @domenipavec I would disable the client to server ping by default on the next release. The server to client ping is better. But you are right, let's store the client to server ping if there is any and skip those clients if the last ping is recent.

server/client_status_check.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (t *ClientsStatusCheckTask) PingClients(jobs <-chan *clients.Client, results chan<- bool) {
+	for j := range jobs {
+		ok, response, rtt, err := comm.PingClientWithTimeout(j.Connection, t.pingTimeout)
+		//t.log.Debugf("ok=%s, error=%s, response=%s", ok, err, response)
+		var now = time.Now()
+		//Old clients cannot respond properly to a ping request yet
+		if !ok && err == nil && string(response) == "unknown request" {
+			t.log.Debugf("ping to %s [%s] succeeded in %s. client < 0.8.2", j.Name, j.ID, rtt)
+			j.LastHeartbeatAt = &now
```
domenipavec, line 83: You have a bit of inconsistent behaviour. On client ping you're saving the heartbeat to the database, but not here. Should this be saved using `SetLastHeartbeat` in client service?

server/client_status_check.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (t *ClientsStatusCheckTask) Run(ctx context.Context) error {
+	timerStart := time.Now()
+	var dueClients []*clients.Client
+	var confirmedClients = 0
+	var now = time.Now()
+	// Shorten the threshold aka make heartbeat older than it is because the ping response is stored after this check.
+	// Clients would get checked only every second time otherwise.
+	t.th = t.th - 10*time.Second
```
domenipavec, line 36: Won't this shorten the threshold on every run and you'll eventually end up with negative thresholds?

server/config_deprecated_test.go
```diff
@@ -0,0 +1,18 @@
+package chserver
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestServerConfigReplaceDeprecated(t *testing.T) {
+	cfg := ServerConfig{
```
domenipavec, line 10: This test would be more self contained and easier to read if you declared a `MockConfig` in this file instead of using actual `ServerConfig`.

server/server.go
```diff
@@ -243,13 +246,40 @@ func (s *Server) Run() error {
 		return err
 	}
 
+	// Look for deprecated server config settings
+	var repl map[string]string
+	var err error
+	s.config.Server, repl, err = ServerConfigReplaceDeprecated(s.config.Server)
```
domenipavec, line 252: I think this should be called from `config.ParseAndValidate` before doing the validation, so that the correct fields are validated.

server/server.go
```diff
@@ -243,13 +246,40 @@ func (s *Server) Run() error {
...
+		go scheduler.Run(ctx, s.Logger, clients.NewCleanupTask(s.Logger, s.clientListener.clientService.repo), s.config.Server.PurgeDisconnectedClientsInterval)
+		s.Infof("Task to purge disconnected clients will run with interval %v", s.config.Server.PurgeDisconnectedClientsInterval)
+	} else {
+		s.Debugf("Task to purge disconnected clients disabled")
 	}
 
+	// We must validate here because the early config validation does not allow logging and continue with overwrites.
+	if s.config.Server.CheckClientsConnectionInterval < CheckClientsConnectionIntervalMinimum {
+		s.config.Server.CheckClientsConnectionInterval = CheckClientsConnectionIntervalMinimum
+		s.Errorf("'check_clients_status_interval' too fast. Using the minimum possible of %s", CheckClientsConnectionIntervalMinimum)
```
domenipavec, line 272: We should try to keep config validation in one place. We already have a case of logging and continuing in client:
https://github.com/cloudradar-monitoring/rport/blob/master/client/config.go#L309

Another option would be to have a custom error type for continuation errors and then use type assertion to check whether to log and continue or abort:
https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go

server/client_listener.go
```diff
@@ -399,6 +399,12 @@ func (cl *ClientListener) handleSSHRequests(clientLog *logger.Logger, clientID s
...
+			now := time.Now()
+			err := cl.clientService.SetLastHeartbeat(clientID, &now)
```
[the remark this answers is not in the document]
thorstenkramm: When a new client connects for the first time, it's nil.
I would not consider the initial connection is a heartbeat.

server/client_status_check.go
thorstenkramm: I admit, saving to the database immediately is superfluous.

client/client.go
```diff
@@ -53,11 +57,18 @@ type Client struct {
 //NewClient creates a new client instance
 func NewClient(config *ClientConfigHolder, filesAPI files.FileAPI) *Client {
 	ctx := context.Background()
-
+	// Generate a session id that will not change while the client is running
+	// This allows the server to resume sessions.
+	sessionID, err := random.UUID4()
+	if err != nil {
+		log.Fatalf("Failed to create initial session id: %s", err)
```
domenipavec, line 64: It would be better to return error and use `Fatal` in main then exiting here.

client/client.go
```diff
@@ -129,7 +141,14 @@ func (c *Client) keepAliveLoop() {
 	for c.running {
 		time.Sleep(c.configHolder.Connection.KeepAlive)
 		if c.sshConn != nil {
-			_, _, _ = c.sshConn.SendRequest(comm.RequestTypePing, true, nil)
+			//ok, response, err := c.sshConn.SendRequest(comm.RequestTypePing, true, nil)
+			ok, _, rtt, err := comm.PingClientWithTimeout(c.sshConn, c.configHolder.Connection.KeepAliveTimeout)
```
domenipavec, line 145: Maybe you should drop `Client` from the function name if it's also used here.

share/logger/logger_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLogger(t *testing.T) {
+	logfile := t.TempDir() + "/test.log"
+	l, err := os.OpenFile(logfile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0444)
+	assert.NoError(t, err, "error creating log file")
+	defer l.Close()
+	logger := NewLogger("test", LogOutput{File: l}, LogLevelDebug)
```
domenipavec, line 15: You can make tests more readable if you put empty lines between setup, execution and assertion phases. In this case after lines 15 and 18.

share/logger/logger_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLogger(t *testing.T) {
+	logfile := t.TempDir() + "/test.log"
+	l, err := os.OpenFile(logfile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0444)
+	assert.NoError(t, err, "error creating log file")
```
domenipavec, line 13: You should usually use `require` for checking errors in tests instead of `assert`. `require` aborts the test on error, and if the file is not created there's no point in continuing.

share/logger/memlogger_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestMemLogger(t *testing.T) {
+	mLog := NewMemLogger()
+	mLog.Debugf("Debug %s", "Debug")
+	mLog.Infof("Info %s", "Info")
+	mLog.Errorf("Error %s", "Error")
+	logfile := t.TempDir() + "/test.log"
+	l, err := os.OpenFile(logfile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0444)
+	assert.NoError(t, err, "error creating log file")
```
domenipavec, line 17: Same

domenipavec (approved):

domenipavec (approved):

PR #644, opened 2022-07-20
[DEV-2607] Extend tunnels with name field

morestatic:
Added the name field to the active tunnel model (called Remote)

Notes:
- Refactored the ClientService into the ClientService interface and a ClientServiceProvider implementation
- Add API unit test to check for the tunnel name field (using a mock based on the ClientService interface)

### DIFF
--- a/server/api_handler_clients_test.go
+++ b/server/api_handler_clients_test.go
@@ -385,25 +385,19 @@ func TestHandlePutTunnelWithName(t *testing.T) {
 	connMock.ReturnResponsePayload = []byte("{ \"IsAllowed\": true }")
 
 	testCases := []struct {
-		Name           string
-		URL            string
-		TunnelName     string
-		ExpectedStatus int
-		ExpectedJSON   string
+		Name         string
+		URL          string
+		ExpectedJSON string
 	}{
 		{
-			Name:           "With Name",
-			URL:            "/api/v1/clients/client-1/tunnels?scheme=ssh&acl=127.0.0.1&local=0.0.0.0%3A3390&remote=0.0.0.0%3A22&name=TUNNELNAME&check_port=0",
-			ExpectedStatus: http.StatusOK,
-			TunnelName:     "TUNNELNAME",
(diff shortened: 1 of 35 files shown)


thorstenkramm (approved):
@domenipavec Please have a look as well.

server/api_handler_clients_test.go
```diff
@@ -343,3 +350,100 @@ func TestGetCorrespondingSortFuncError(t *testing.T) {
...
+
+func TestHandlePutTunnelWithName(t *testing.T) {
+	connMock := test.NewConnMock()
+	connMock.ReturnOk = true
+	connMock.ReturnResponsePayload = []byte("{ \"IsAllowed\": true }")
+
+	testCases := []struct {
+		Name           string
+		URL            string
+		TunnelName     string
```
domenipavec, line 390: This doesn't seem to be used.

server/server.go
```diff
@@ -309,6 +309,10 @@ func (s *Server) Close() error {
 	return wg.Wait()
 }
 
+func (s *Server) GetClientService() ClientService {
```
domenipavec, line 312: This is not used?

domenipavec (dismissed):

server/server.go
morestatic: ah good catch. yes, not required. i've removed it.

server/api_handler_clients_test.go
morestatic: I can't quite see (or not quite getting) what you're referencing. But I can see that the ExpectedStatus isn't really being used. Is that what you mean? I'll tidy the ExpectedStatus up but let me know if something else? The use of a test table is borderline. Do you mean that?

domenipavec: I meant that `TunnelName` property is set in `testCases` but not used in the actual tests.

morestatic: you're right. i really couldn't see that one. i think for a while i was going to Sprintf it into the strings but decided that was over the top. ok. removed and also removed the ExpectedStatus use.

thorstenkramm (approved):

PR #657, opened 2022-08-01
Feature/dev 2525 group permissions

thorstenkramm:
@domenipavec I did some rework to
* Let people pass, if group permissions are not activated or not supported. Otherwise, all users expect Admins would be blocked.
* MySQL behaves slightly different. Json unmarshalling needs a MySQL variant and affected rows are always 0. Using `REPLACE INTO` based on a unique index works for MySQL and SQLite.
* Return the effective user permissions, so the frontend can hide areas.


server/api_listener.go
```diff
@@ -494,6 +495,17 @@ func (al *APIListener) wsAuth(f http.Handler) http.HandlerFunc {
 			return
 		}
 
+		if reqPermission != "" && al.userService.GetProviderType() == enums.ProviderSourceDB {
```
domenipavec, line 498: I think you should be able to just use `permissionsMiddleware` after `wsAuth` middleware instead of duplicating the code.

server/api_router.go
```diff
@@ -152,9 +152,9 @@ func (al *APIListener) initRouter() {
 
 	// web sockets
 	// common auth middleware is not used due to JS issue https://stackoverflow.com/questions/22383089/is-it-possible-to-use-bearer-authentication-for-websocket-upgrade-requests
-	api.HandleFunc("/ws/commands", al.wsAuth(http.HandlerFunc(al.handleCommandsWS))).Methods(http.MethodGet)
-	api.HandleFunc("/ws/scripts", al.wsAuth(http.HandlerFunc(al.handleScriptsWS))).Methods(http.MethodGet)
-	api.HandleFunc("/ws/uploads", al.wsAuth(http.HandlerFunc(al.handleUploadsWS))).Methods(http.MethodGet)
+	api.HandleFunc("/ws/commands", al.wsAuth(http.HandlerFunc(al.handleCommandsWS), users.PermissionCommands)).Methods(http.MethodGet)
+	api.HandleFunc("/ws/scripts", al.wsAuth(http.HandlerFunc(al.handleScriptsWS), users.PermissionScripts)).Methods(http.MethodGet)
+	api.HandleFunc("/ws/uploads", al.wsAuth(http.HandlerFunc(al.handleUploadsWS), users.PermissionUploads)).Methods(http.MethodGet)
```
domenipavec, line 157: Since you're protecting the uploads status endpoint, you should also protect actual upload handler `al.handleFileUploads`

server/api_handler_me.go
```diff
@@ -24,12 +24,37 @@ func (al *APIListener) handleGetMe(w http.ResponseWriter, req *http.Request) {
 		al.jsonErrorResponseWithTitle(w, http.StatusNotFound, "user not found")
 		return
 	}
-
-	me := UserPayload{
+	var me = UserPayload{
```
domenipavec, line 27: Using `:=` instead of `var` is more common in go when you are also assigning the value.

server/api_handler_me.go
```diff
@@ -24,12 +24,37 @@ func (al *APIListener) handleGetMe(w http.ResponseWriter, req *http.Request) {
...
 		Username:    user.Username,
 		Groups:      user.Groups,
 		TwoFASendTo: user.TwoFASendTo,
 	}
+	var eup = EffectiveUserPermissions{}
+	eup.Permissions = map[string]bool{}
```
domenipavec, line 33: You should do this in one step. Also using `make` is common practice when initializing empty maps.
```
eup := EffectiveUserPermissions{
    Permissions: make(map[string]bool),
}
```

api-doc/openapi/components/schemas/UserGet.yaml
```diff
@@ -13,3 +13,37 @@ properties:
     description: >-
       Holds either the email or the pushover token of the user. It's used to
       deliver 2FA token to user when 2FA is enabled
+  effective_user_permissions:
+    type: object
+    description: The effective user right inhereted from the group permissions
+    properties:
+      note:
```
domenipavec, line 20: I think having strings `note` is a bad practice in API. You should just have a `bool` value to indicate whether group permissions are supported and frontend can display approparite message.

server/api_handler_me.go
```diff
@@ -24,12 +24,37 @@ func (al *APIListener) handleGetMe(w http.ResponseWriter, req *http.Request) {
...
+	for _, perm := range users.AllPermissions {
+		eup.Permissions[perm] = false
+	}
+	if !al.userService.SupportsGroupPermissions() {
+		eup.Note = "effective_user_permissions: not supported by api auth provider"
+		for _, perm := range users.AllPermissions {
+			eup.Permissions[perm] = true
+		}
+	} else {
+		for _, groupName := range user.Groups {
```
domenipavec, line 43: I think it would be better to have a `UserPermissions` method on `userService` that returns a `Permissions` object instead of implementing it here.

server/api_handler_me.go
```diff
@@ -24,12 +24,37 @@ func (al *APIListener) handleGetMe(w http.ResponseWriter, req *http.Request) {
...
-	me := UserPayload{
+	var me = UserPayload{
 		Username:    user.Username,
 		Groups:      user.Groups,
 		TwoFASendTo: user.TwoFASendTo,
 	}
+	var eup = EffectiveUserPermissions{}
+	eup.Permissions = map[string]bool{}
+	for _, perm := range users.AllPermissions {
+		eup.Permissions[perm] = false
```
domenipavec, line 35: You should user `Permissions` object to avoid this loop.

server/api_listener.go
thorstenkramm: wsAuth has been created because browsers cannot send headers when switching to ws. I tried the middleware, and it's returning 401 always.

domenipavec: But `wsAuth` uses the same `api.WithUser` to set user on the context as the regular auth. So `permissionMiddleware` should work fine if it's set inside `wsAuth`. If it doesn't work we should figure out why.

server/api/users/permissions.go
```diff
@@ -49,6 +51,14 @@ func (permissions Permissions) All() map[string]bool {
 	return result
 }
 
+func CreateDefaultPermissions(set bool) map[string]bool {
```
domenipavec, line 54: You should use the actual permissions object. It's designed in a way that an empty map will serialize to correct json with all permissions set to false. So this function with `set=false` would not be needed. And for the `set=true` case you should just construct it the way it's done in admin group. So this function is not needed.

server/api_listener.go
```diff
@@ -494,6 +496,17 @@ func (al *APIListener) wsAuth(f http.Handler) http.HandlerFunc {
 			return
 		}
 
+		if reqPermission != "" && al.userService.SupportsGroupPermissions() {
```
domenipavec, line 499: Even if we're not using bearer auth for ws, it should still be possible to use `permissionsMiddleware` inside `wsAuth`.

domenipavec (approved):

PR #658, opened 2022-08-02
[DEV-2537, 2631, 2632] Use tags for targeting

morestatic:
Allows command, scripts, scheduled jobs and file upload features to use tags for targeting clients for execution related tasks.

Note:
This has become a rather heavy set of changes with potential functional impact. Strongly recommend heavy manual testing for the various execution scenarios.


server/api_handler_schedules.go
```diff
@@ -37,11 +38,15 @@ func (al *APIListener) handlePostSchedules(w http.ResponseWriter, req *http.Requ
 		al.jsonError(w, err)
 		return
 	}
-	orderedClients, _, err := al.getOrderedClients(ctx, scheduleInput.Details.ClientIDs, scheduleInput.Details.GroupIDs, true /* allowDisconnected */)
+
+	var orderedClients []*clients.Client
+
+	orderedClients, _, isBadRequest, errTitle, err := al.getOrderedClientsWithValidation(ctx, &scheduleInput, 2)
```
domenipavec, line 44: Schedules don't have a separate endpoint for single client execution, so we should definitely allow only 1 client here.

server/api_handler_schedules.go
```diff
@@ -85,11 +90,15 @@ func (al *APIListener) handleUpdateSchedule(w http.ResponseWriter, req *http.Req
...
```
domenipavec, line 96: Same

server/api_handler_schedules.go
```diff
@@ -37,11 +38,15 @@ func (al *APIListener) handlePostSchedules(w http.ResponseWriter, req *http.Requ
 		al.jsonError(w, err)
 		return
 	}
-	orderedClients, _, err := al.getOrderedClients(ctx, scheduleInput.Details.ClientIDs, scheduleInput.Details.GroupIDs, true /* allowDisconnected */)
+
+	var orderedClients []*clients.Client
```
domenipavec, line 42: Why is this a separate definition, it would get defined below anyway?

server/api_handler_commands.go
```diff
@@ -290,23 +291,13 @@ func (al *APIListener) handlePostMultiClientCommand(w http.ResponseWriter, req *
 		return
 	}
 
-	orderedClients, groupClientsCount, err := al.getOrderedClients(ctx, reqBody.ClientIDs, reqBody.GroupIDs, false /* allowDisconnected */)
+	orderedClients, _, isBadRequest, errTitle, err := al.getOrderedClientsWithValidation(ctx, &reqBody, 2)
```
domenipavec, line 294: Please make `2` a constant with descriptive name.
Same for other places where you use constant numbers.

server/api_handler_commands_test.go
```diff
@@ -719,3 +739,1238 @@ func TestHandlePostMultiClientCommand(t *testing.T) {
 		})
 	}
 }
+
+func TestHandlePostMultiClientCommandWithGroupIDs(t *testing.T) {
```
domenipavec, line 743: I haven't dived deep in these tests, but they look pretty similar. Would it be possible to have them in the same test as separate testcases?

server/api/jobs/jobs.go
```diff
@@ -101,6 +101,10 @@ func NewSqliteProvider(db *sqlx.DB, log *logger.Logger) *SqliteProvider {
 	}
 }
 
+func (p *SqliteProvider) GetDB() (db *sqlx.DB) {
```
domenipavec, line 104: Is this used anywhere?

server/api_helper_clients.go
```diff
@@ -2,14 +2,65 @@ package chserver
...
+	GetClientIDs() (ids []string)
+	GetGroupIDs() (ids []string)
+	GetClientTags() (clientTags *models.JobClientTags)
+	GetTags() (tags []string)
+}
+
+func (al *APIListener) getOrderedClientsWithValidation(
+	ctx context.Context,
+	params TargetingParams,
+	minClients int,
```
domenipavec, line 31: I question having this configurable, especially if it's not actually used in case of `tags` targeting. Maybe we should just change the behavior and always check for minimum of 1 client?

server/api_helper_commands.go
```diff
@@ -146,13 +138,19 @@ func (al *APIListener) handleCommandsExecutionWS(
...
+
 				if !success {
 					if multiJob.AbortOnErr {
 						uiConnTS.Close()
 						return
 					}
 					continue
 				}
+
+				if al.insecureForTests {
```
domenipavec, line 150: `insecureForTests` has a very specific function - disabling authentication for tests. I wouldn't use it here. If you require different behavior in tests here you should probably use a different flag or modify the mocks so that it's not needed.

server/cgroups/sqlite.go
```diff
@@ -24,11 +20,7 @@ type SqliteProvider struct {
 	db *sqlx.DB
 }
 
-func NewSqliteProvider(dbPath string, dataSourceOptions sqlite.DataSourceOptions) (*SqliteProvider, error) {
-	db, err := sqlite.New(dbPath, client_groups.AssetNames(), client_groups.Asset, dataSourceOptions)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create client_groups DB instance: %v", err)
-	}
+func NewSqliteProvider(db *sqlx.DB) (*SqliteProvider, error) {
```
domenipavec, line 23: Why change this? It makes initialization less self contained and puts even more code in server.go initialization.
I know we do it this way for some other sqlite providers, but that's because we share `sqlx.DB` between multiple providers. Here that's not needed.

server/server.go
```diff
@@ -356,3 +367,11 @@ func (m *jobResultChanMap) Get(jobID string) chan *models.Job {
 	defer m.mu.RUnlock()
 	return m.m[jobID]
 }
+
+func (m *jobResultChanMap) GetAllKeys() (jobIDs []string) {
```
domenipavec, line 371: Is this used anywhere?
Since the description of the struct is that it is thread safe, this should use the mutex.

server/clients/cr.go
```diff
@@ -91,6 +92,59 @@ func (s *ClientRepository) Delete(client *Client) error {
 	return nil
 }
 
+func (s *ClientRepository) GetClientsByTag(tags []string, operator string, allowDisconnected bool) (matchingClients []*Client, err error) {
+	var availableClients []*Client
+	if allowDisconnected {
+		availableClients, err = s.GetAll()
```
domenipavec, line 98: I know this error is technically handled, but it makes the code less readable and harder to change. Please use `if err != nil { return nil, err }`

server/clients/cr.go
```diff
@@ -91,6 +92,59 @@ func (s *ClientRepository) Delete(client *Client) error {
...
+	} else {
+		availableClients = s.GetAllActive()
+	}
+	if err == nil && strings.ToUpper(operator) == "AND" {
```
domenipavec, line 102: You can use `strings.EqualFold`

server/clients/cr.go
```diff
@@ -91,6 +92,59 @@ func (s *ClientRepository) Delete(client *Client) error {
...
+
+func findMatchingANDClients(availableClients []*Client, tags []string) (matchingClients []*Client) {
+	matchingClients = make([]*Client, 0, 64)
+	for _, cl := range availableClients {
+		clientTags := cl.Tags
+		foundCount := 0
+		for _, tag := range tags {
+			for _, clTag := range clientTags {
+				if tag == clTag {
+					foundCount++
```
domenipavec, line 119: I'm not sure `foundCount` is a good measure as I don't think we validate that there's no duplicate tags. You should just add a boolean found for each tag and check that.

server/clients/cr.go
```diff
@@ -91,6 +92,59 @@ func (s *ClientRepository) Delete(client *Client) error {
...
+			}
+		}
+		if foundCount == len(tags) {
+			matchingClients = append(matchingClients, cl)
+		}
+	}
+	return matchingClients
+}
+
+func findMatchingORClients(availableClients []*Client, tags []string) (matchingClients []*Client) {
```
domenipavec, line 131: It would be nice to have unit tests for these 2 functions since the logic is non trivial.

domenipavec (changes requested):
Reviewing PRs this big is very time consuming. I would appreciate it if you split your code changes into smaller PRs in the future.

I like all the tests you added :+1:

morestatic:
hi @domenipavec, 

thanks for the detailed review. overall, i think nearly all the comments will be accepted. 

yeah, the PR did get too big. agree that the goal should be to keep them smaller and easier to review.

i did talk with @thorstenkramm about refactoring the test code, and we decided to defer for now but it is an outstanding task that needs to be done. i saw your comment about the same test with separate test cases. i think we should take care not to create 'uber' tests that become awkward to deal with. however, there is lots of duplicated test code (and cases) that we need to de-duplicate. i'd prefer to do this via shared functions that are called via more tests rather than making one bigger test. when we decide to do the refactoring happy to discuss the approach though and maybe we can have a call to agree the best way.

noticed your comment about a single client for scheduled jobs. sounds like i don't quite understand how scheduled jobs work. will ask @thorstenkramm to clarify in the related comment. 

ok, thanks again for digging in for the review. will get on with addressing the individual points.

server/api_handler_schedules.go
morestatic: @thorstenkramm (or maybe @domenipavec) could you clarify the requirement? not sure i understand how the feature works and what we need. the api-docs say client_ids and group_ids are supported, so how and where does the 1 client constraint apply?

server/api_handler_schedules.go
morestatic: done

server/api_handler_commands.go
morestatic: done

server/api/jobs/jobs.go
morestatic: yep, used in the creation of a `scheduleManager` for test dependency purposes. see below.
https://github.com/cloudradar-monitoring/rport/blob/85cc222f18907e97504727f556c02710e7706dd1/server/api_handler_commands_test.go#L1971
btw, i think how we deal with test dependencies is something we can cover when discussing the test code refactoring. you're much more familar the code base, so might prefer a different approach. mine was somewhat brute force. I'll add a TODO in the code, so that we find to discuss later.

server/server.go
morestatic: yes, again this is something introduced for testing purposes. when testing the WS api, there isn't any available feedback for the test. the multijob id is written by the client listener to directly to the WS, and the client listener isn't instantiated as part of the api listener used for testing. therefore, the test introspects the job results chan keys to find the multijob id, which is then used to find a count of created jobs for checking expectations in the test. i can't say i feel confident about the approach (due to a potential race before the job chan is removed) but it does seem to work (due to the chan interaction sequencing in the test scenario). any thoughts?
https://github.com/cloudradar-monitoring/rport/blob/85cc222f18907e97504727f556c02710e7706dd1/server/api_handler_commands_test.go#L1358

even though only used for testing, agree that it should have the mutex protection. will also add a comment that used for test purposes.

server/upload.go
```diff
@@ -315,22 +347,30 @@ func (al *APIListener) uploadRequestFromRequest(req *http.Request) (*UploadReque
...
+	clientTags = &models.JobClientTags{}
+	err = json.Unmarshal([]byte(jsonTags[0]), clientTags)
+	if err != nil {
+		return nil, err
 	}
 
+	return clientTags, nil
+}
+
+func validateUploadRequest(ur *UploadRequest) error {
```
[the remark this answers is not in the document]
morestatic: done

server/clients/cr.go
```diff
@@ -32,10 +33,10 @@ type User interface {
 // keepDisconnectedClients is a duration to keep disconnected clients. If a client was disconnected longer than a given
 // duration it will be treated as obsolete.
 func NewClientRepository(initClients []*Client, keepDisconnectedClients *time.Duration, logger *logger.Logger) *ClientRepository {
-	return newClientRepositoryWithDB(initClients, keepDisconnectedClients, nil, logger)
+	return NewClientRepositoryWithDB(initClients, keepDisconnectedClients, nil, logger)
 }
 
-func newClientRepositoryWithDB(initClients []*Client, keepDisconnectedClients *time.Duration, provider ClientProvider, logger *logger.Logger) *ClientRepository {
+func NewClientRepositoryWithDB(initClients []*Client, keepDisconnectedClients *time.Duration, provider ClientProvider, logger *logger.Logger) *ClientRepository {
```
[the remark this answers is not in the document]
morestatic: again, this was related to creating test dependencies. the `NewClientRepositoryWithDB` function is called in two separate packages (server and server/clients), so can't remain private to one package. however, i'm not sure we're really using the function as initially planned (by me) and was planning to revisit as part of the refactoring. can we come back to this function then? and for now, i'll add a TODO comment?

server/clients/cr.go
morestatic: agreed. done.

server/clients/cr.go
morestatic: done

server/api_helper_commands.go
morestatic: yes, understand. i was just following the pattern previously implemented. 
https://github.com/cloudradar-monitoring/rport/blob/85cc222f18907e97504727f556c02710e7706dd1/server/api_helper_commands.go#L150
agreed about potentially using a new flag or a different mechanism entirely. again, i think we should wait until the test refactor to address? originally, i tried to do something fancy and went down a mocking rabbit hole that was too ambitious for my novice level of understanding of the code base and behavior, so reverted after burning a few hours. will add TODO comments for the flag use.

server/api/jobs/schedule/manager.go
```diff
@@ -95,6 +88,18 @@ func New(ctx context.Context, logger *logger.Logger, db *sqlx.DB, jobRunner JobR
 	return m, nil
 }
 
+func NewManager(jobRunner JobRunner, db *sqlx.DB, logger *logger.Logger, runRemoteCmdTimeoutSec int) (m *Manager) {
```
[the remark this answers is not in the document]
morestatic: it was related to test dependencies, i wanted to avoid the test spinning up the cron behaviour, which may not have had a valid instantiation for testing purpose and could have crashed. it is a little difficult discovering all the dependency requirements for the tests, so was keen to minimise additional component use.

server/api/jobs/schedule/manager.go
```diff
@@ -233,14 +238,6 @@ func (m *Manager) validate(s *Schedule) error {
 		}
 	}
 
-	if len(s.Details.ClientIDs) == 0 && len(s.Details.GroupIDs) == 0 {
```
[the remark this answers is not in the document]
morestatic: i did think about this but decided against it, as the old behaviour would soon be forgotten and the comment would have become stale. ideally the PR could have described how the changes impacted the checks. on a larger project, I think the PR notes could have been important.

server/api_handler_schedules.go
morestatic: done

server/api_helper_clients.go
morestatic: i was a little puzzled why only 1 client was an error? do you know/remember why? it seemed fair that particularly a group might only have a single member at times and that would be valid when targeting group?

server/api_handler_schedules.go
domenipavec: @morestatic Maybe my comment was unclear. I meant that the argument for min number of clients should be `1`, so that 1 or more clients are allowed, not only 2 or more.

server/api_helper_clients.go
domenipavec: I was not the one implementing the feature so I'm not sure. But I think it was meant to encourage users to use single client command execution when there's only 1 client.
I think if @thorstenkramm agrees we can remove the error when there's only 1 client.

server/clients/cr.go
morestatic: done

server/clients/cr.go
morestatic: done

server/cgroups/sqlite.go
morestatic: again, this was another test dependency mod. we aren't only concerned with the server initialisation but also need to support test setup / dependencies. the usual pattern seems to be to provide a slightly higher level version of the function (for the server initialisation) that then calls the slightly lower level version (which can then also be used for test setup/dependencies). Whether internal or external depends on where the tests are.

server/api_helper_clients.go
morestatic: i think consensus is forming that we only need a check for at least 1 client and that the 2 clients check probably isn't required anymore. but i'll confirm with @thorstenkramm.

server/api_handler_commands_test.go
morestatic: As mentioned previously, we need to refactor the tests. There are multiple opportunities for them to be improved. In principle, I'd like to stick with tests per api, but it does get interesting when a number of the apis shared the same validation. now we have the shared validation fn, we should probably have unit tests for that individual fn and then less in the api tests. we can do this when we do the test refactoring.

server/api_handler_schedules.go
morestatic: done

server/api_helper_clients.go
```diff
@@ -2,14 +2,65 @@ package chserver
...
+	ErrRequestIncludesMultipleTargetingParams = errors.New("multiple targeting options are not supported. Please specify only one")
+	ErrRequestMissingTargetingParams          = errors.New("please specify targeting options, such as client ids, groups ids or tags")
+	ErrMissingTagsInMultiJobRequest           = errors.New("please specify tags in the tags list")
+)
+
+type TargetingParams interface {
+	GetClientIDs() (ids []string)
+	GetGroupIDs() (ids []string)
+	GetClientTags() (clientTags *models.JobClientTags)
+	GetTags() (tags []string)
```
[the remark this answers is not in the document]
morestatic: done

server/clients/test_helpers.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/cloudradar-monitoring/rport/db/migration/clients"
+	"github.com/cloudradar-monitoring/rport/db/sqlite"
+)
+
+var DataSourceOptions = sqlite.DataSourceOptions{WALEnabled: false}
+
+func NewFakeClientProvider(t *testing.T, exp *time.Duration, cs ...*Client) *SqliteProvider {
```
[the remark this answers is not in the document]
morestatic: This is another function used for test setup.

server/api_handler_schedules.go
morestatic: @domenipavec ok. yes, we're gonna move to checking for a single client only. that's done now.

server/api_helper_clients.go
```diff
@@ -71,20 +149,87 @@ func (al *APIListener) getOrderedClients(
...
+		return fmt.Errorf(fmt.Sprintf("At least %d client should be specified.", minClients))
+	}
+	return nil
+}
+
+func hasClientTags(params TargetingParams) (has bool) {
+	return params.GetTags() != nil
+}
+
+func (al *APIListener) makeJSONErr(w http.ResponseWriter, err error, errTitle string, isBadRequest bool) {
```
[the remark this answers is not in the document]
morestatic: ok. yes, tried to tidy up the error handling and move everything to APIErrors. we might need one more iteration to finalise but not too far away now. please check the latest commits.

server/api_helper_clients.go
```diff
@@ -71,20 +151,98 @@ func (al *APIListener) getOrderedClients(
...
+		return errors2.APIError{
+			Err:        errors.New("at least 1 client should be specified"),
+			HTTPStatus: http.StatusBadRequest,
+		}
+	}
+	return nil
+}
+
+func validateClientTagsTargeting(orderedClients []*clients.Client) (err error) {
+	if orderedClients == nil || len(orderedClients) < minClientsForTargeting {
```
domenipavec, line 237: `len` of nil slice is 0, so you don't have to make 2 separate checks.

server/api_responses.go
```diff
@@ -65,3 +65,12 @@ func (al *APIListener) jsonErrorResponseWithError(w http.ResponseWriter, statusC
 	}
 	al.writeJSONResponse(w, statusCode, api.NewErrAPIPayloadFromMessage("", title, detail))
 }
+
+func (al *APIListener) jsonErrorResponseWithAPIError(w http.ResponseWriter, err error) {
```
domenipavec, line 69: We already have `jsonError` method that should correctly handle `APIError`, please don't add additional method.

server/clients/cr.go
```diff
@@ -91,6 +93,72 @@ func (s *ClientRepository) Delete(client *Client) error {
...
+		for _, tag := range tags {
+			// ensure map has an entry for each desired unique tag
+			matchingTags[tag] = false
+			for _, clTag := range clientTags {
+				if tag == clTag {
+					matchingTags[tag] = true
+					continue
+				}
+			}
+		}
```
domenipavec, line 130: Here's how I would simplify this code:
```
foundAllTags := true
for _, tag := range tags {
  foundTag := false
  for _, clTag := range clientTags {
    if tag == clTag {
      foundTag = true
      break
    }
  }
  if !foundTag {
    foundAllTags = false
    break
  }
}
if foundAllTags {
  matchingClients = append(matchingClients, cl)
}
```
No need for maps and counts.

server/api/jobs/model.go
```diff
@@ -36,9 +39,24 @@ func (req *MultiJobRequest) GetClientTags() (clientTags *models.JobClientTags) {
 	return req.ClientTags
 }
 
-func (req *MultiJobRequest) GetTags() (ids []string) {
-	if req.ClientTags == nil {
-		return nil
+// TODO: add some unit tests. not high priority but good to get done.
+func MakeClientTagsAsString(jobTags *models.JobClientTags) (clientTags string) {
```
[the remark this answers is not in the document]
morestatic: i did think about it but what happens when the jobClientTags are nil? when I tried I got `<nil>` in the output. did I miss something, that could allow for rendering a nicer nil value?

server/api_helper_clients.go
morestatic: old habits die hard :)

done.

server/clients/cr.go
morestatic: ok. yes, this is better. thanks.

done.

server/api_responses.go
morestatic: done.

server/api/jobs/model.go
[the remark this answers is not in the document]
morestatic: Haven't actually used nil receivers before, so thought i'd try. Yes, works fine. So moved this helper function to a stringer, as suggested. It feels better.

domenipavec (approved):

PR #665, opened 2022-08-11
Feature/dev 2507 connection analysis

thorstenkramm:


client/clientutils/connetion_error_hints.go
```diff
@@ -0,0 +1,60 @@
+package clientutils
```
domenipavec, line 1: Anything with utils in the name is too generic for a package name in my opinion. I would just put these functions in the main client package.

client/clientutils/connetion_error_hints.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func DetectTransparentProxy(serverURL string) (proxy string, allHeaders string, error error) {
+	var client = &http.Client{
+		Timeout: time.Second * 10,
+	}
+	res, err := client.Head(strings.Replace(serverURL, "ws", "http", 1))
+	if err != nil {
+		if os.IsTimeout(err) {
+			// timeout
```
domenipavec, line 44: This comment is not very helpful, that this is the case of timeout is obvious from the if condition.

client/clientutils/connetion_error_hints.go
```diff
@@ -0,0 +1,60 @@
...
+			return "", "", fmt.Errorf("error on http HEAD request: %s", err)
```
domenipavec, line 45: Why even have this if condition, it's handled the same?

client/client.go
```diff
@@ -326,7 +328,8 @@ func (c *Client) connect(server string) (*sshClientConn, error) {
 	}
 	wsConn, _, err := d.Dial(server, c.configHolder.Connection.HTTPHeaders)
 	if err != nil {
-		return nil, retryableError{err}
+		//return nil, retryableError{err}
```
domenipavec, line 331: Remove

domenipavec (approved):

PR #667, opened 2022-08-15
a watchdog integration

thorstenkramm:


client/watchdog_test.go
```diff
@@ -0,0 +1,57 @@
...
+		t.Run(tc.state, func(t *testing.T) {
+			w.Ping(tc.state, tc.msg)
+			s, err := os.ReadFile(stateFile)
+			require.NoError(t, err)
+			var ws watchdogState
+			err = json.Unmarshal(s, &ws)
+			require.NoError(t, err)
+			assert.Equal(t, tc.expected, ws.LastState)
+			assert.Equal(t, tc.msg, ws.LastMessage)
+			assert.NotEqual(t, lastUpdate, ws.LastUpdate)
```
domenipavec, line 53: `NotEqual` is a very vague assertion. I would assert that the update time is within some duration of now with `assert.WithinDuration`.

client/watchdog.go
```diff
@@ -0,0 +1,110 @@
...
+	WatchdogStateConnected    = "connected"
+	WatchdogStateInit         = "initialized"
+
+	// SdNotifyWatchdog tells the service manager to update the watchdog
+	// timestamp for the service.
+	sdNotifyWatchdog = "WATCHDOG=1"
+)
+
+type Watchdog struct {
+	enabled    bool
```
domenipavec, line 26: Nit.: This implementation is fine. 
But you could just return `nil` watchdog when it is not enabled and check `w == nil` in `Ping`. That way you avoid having an enabled field.

client/watchdog.go
```diff
@@ -0,0 +1,110 @@
...
+		w.logger.Errorf(err.Error())
+	}
+}
+
+func (w *Watchdog) sdNotify() error {
+	if w.socketAddr.Name == "" {
+		return nil
+	}
+
+	conn, err := net.DialUnix(w.socketAddr.Net, nil, w.socketAddr)
```
domenipavec, line 100: Nit.: It might be more efficient to create a connection in `NewWatchdog` and reuse it every time?

domenipavec (approved):
Some small comments, otherwise looks good.

client/watchdog.go
thorstenkramm: But where will the connection be closed?
The `defer conn.Close()` must then be executed by the caller `NewClient()` with an if/else depending on if there is a socket connection or not.

client/watchdog.go
thorstenkramm: Returning nil causes all calls to `Watchdog.Ping()` to fail. Checking if != nil on all calls looks like repetition.

client/watchdog.go
domenipavec: You would save the `conn` on the `Watchdog` object and add a `Close` method that would close the connection if it's not `nil`. You would call `Watchdog.Close` from the `Client.Close`.

client/watchdog.go
domenipavec: You would have to check for `nil` immediately inside `Ping` instead of checking if it's enabled. Calling methods on `nil` pointer receiver is perfectly valid in go.

domenipavec (dismissed):

client/watchdog.go
thorstenkramm: Awesome. Now I have understood.

client/watchdog.go
```diff
@@ -89,18 +93,18 @@ func (w *Watchdog) Ping(state string, msg string) {
...
 		return err
 	}
-	defer conn.Close()
+	return nil
+}
 
-	if _, err = conn.Write([]byte(sdNotifyWatchdog)); err != nil {
-		return err
+func (w *Watchdog) Close() error {
+	if w.socket != nil {
```
domenipavec, line 106: I think you have to check for `w == nil` first, otherwise this might panic.

PR #690, opened 2022-09-12
Added timeout field to command's and script's library

thorstenkramm:


server/api/command/model.go
```diff
@@ -17,10 +17,14 @@ type Command struct {
...
-	Name string   `json:"name" db:"name"`
-	Cmd  string   `json:"cmd" db:"script"`
-	Tags []string `json:"tags" db:"tags"`
+	Name      string   `json:"name" db:"name"`
+	Cmd       string   `json:"cmd" db:"script"`
+	Tags      []string `json:"tags" db:"tags"`
+	TimoutSec int      `json:"timeout_sec" db:"timeout_sec"`
 }
+
+const DefaultTimeoutSec = 60
```
domenipavec, line 30: Nit.: It's common practice to put constants at the top of the file.

server/api/command/sqlite_test.go
```diff
@@ -55,7 +58,7 @@ func TestGetByID(t *testing.T) {
 	val, found, err := dbProv.GetByID(ctx, "1", &query.RetrieveOptions{})
 	require.NoError(t, err)
 	require.True(t, found)
-	assert.Equal(t, demoData[0], *val)
+	assert.Equal(t, demoData[0], *val, "first test")
```
domenipavec, line 61: Nit.: Instead of adding message to asserts, it would be better to wrap each sub-case in `t.Run`

server/script/model.go
```diff
@@ -29,4 +30,7 @@ type InputScript struct {
 	Cwd         string   `json:"cwd" db:"cwd"`
 	Script      string   `json:"script" db:"script"`
 	Tags        []string `json:"tags" db:"tags"`
+	TimoutSec   int      `json:"timeout_sec" db:"timeout_sec"`
 }
+
+const DefaultTimeoutSec = 60
```
domenipavec, line 36: Same

domenipavec (dismissed):

PR #699, opened 2022-09-22
[DEV-2618, DEV-2617, DEV-2643] Prepare rport-plus sso / oauth for release

morestatic:
This PR adds support for the first version of the rport-plus plugin with initial OAuth support (GitHub and Microsoft providers) and basic plugin status capabilities. Scope covers the following:

* Initialization of the plugin from the rportd main.go file. This includes validation of related config options for the plus capabilities from the rportd config file.
* An rport-plus manager component that registers and loads capability interfaces (via plugin init funcs). The Plus manager then caches the capability interfaces and allows rport related functionality to access those interfaces when required.
* Defines 'shared' constants, types, etc used by both rportd and the rport-plus plugin.
* Support for permitted user lists based on the existing api auth mechanisms.
* Support for all users from a single Github organization or Microsoft group to automatically be granted access.
* Full configuration via the rportd config file, including the path to the rport-plus plugin.
* Unit tests.
* A basic OAuth mock capability to allow server testing without using an actual plugin.
* Updated api-doc.
* Updated CI / CD with that includes an rport-plus release asset.
* Update rportd.example.conf with config param notes and examples.

Notes for PR Review

* There is moderate use of partial mocks for testing. This seemed like the most appropriate testing approach but I'm open to alternatives.
* Access to the functionality in the plugin is via a single initialization function, that returns a golang interface to the individual functions of each capability. Initially each capability function was exposed as a separately accessible plugin function. Exposing an interface reduces the surface area of exposed plugin functions and makes testing of capabilities much more straightforward.
* When building rport and rport-plus together, rport-plus expects to reference local rport packages (via the `replace` directive in the rport-plus `go.mod` file).


server/config.go
```diff
@@ -255,6 +257,10 @@ type Config struct {
 	Pushover   PushoverConfig   `mapstructure:"pushover"`
 	SMTP       SMTPConfig       `mapstructure:"smtp"`
 	Monitoring MonitoringConfig `mapstructure:"monitoring"`
+
+	// Config for rport-rplus is here. This is the place to add individual capability configs
+	PlusConfig  *rportplus.PlusConfig `mapstructure:"plus-plugin"`
```
domenipavec, line 262: It might be better if you added `PlusConfig` with `mapstructure:",squash"`, and then separate the oauth and plugin configs in the rportplus code. That way you wouldn't have to add capability configs here.

plus/capabilities/oauth/oauth.go
```diff
@@ -0,0 +1,99 @@
...
+	GitHubOAuthProvider    = "github"
+	MicrosoftOAuthProvider = "microsoft"
+	Auth0OAuthProvider     = "auth0"
+
+	DefaultLoginURI = "/oauth/login"
+)
+
+// Capability is used by rportd to maintain loaded info about the plugin's
+// oauth capability
+type Capability struct {
```
domenipavec, line 70: I'm not sure why a specific `Capability` implementation for each capability is required? Why not just pass `Config` to the `initFn` directly and use `CapabilityEx` in the manager instead of wrapping them in `Capability`?

plus/manager.go
```diff
@@ -0,0 +1,200 @@
...
+	SetProvider(sym plugin.Symbol)
+	GetConfigValidator() (v validator.Validator)
+}
+
+// Manager defines the plus manager behavior
+type Manager interface {
+	InitPlusManager(cfg *PlusConfig, logger *logger.Logger)
+	RegisterCapability(capName string, newCap Capability) (cap Capability, err error)
+	SetCapability(capName string, cap Capability)
+	GetCapability(capName string) (cap Capability)
```
domenipavec, line 46: Are `SetCapability` and `GetCapability` used anywhere except in this package and tests? It seems to me they shouldn't really be part of the public interface.

plus/manager.go
```diff
@@ -0,0 +1,200 @@
...
+
+// ManagerProvider contains a map of all available capabilities and the overall
+// plugin config. The manager is thread safe for reads but not initialization.
+type ManagerProvider struct {
+	Config *PlusConfig
+	logger *logger.Logger
+
+	caps map[string]Capability
+
+	mu sync.RWMutex
```
domenipavec, line 62: Is this mutex supposed to protect `caps`? In that case I would define it just above it so it's clear.

Also you're not using the mutex very consistently. I would define `getCap` and `setCap` methods that use mutex and not use the map directly.

server/plus.go
```diff
@@ -0,0 +1,75 @@
...
+
+var (
+	ErrPlusNotEnabled = errors.New("rport-plus not enabled")
+)
+
+// EnablePlusIfLicensed will initialize a new plus manager and request registration of the desired
+// capabilities
+func EnablePlusIfLicensed(cfg *Config, filesAPI files.FileAPI) (plusManager rportplus.Manager, err error) {
+	if cfg.PlusConfig == nil || (cfg.PlusConfig != nil && cfg.PlusConfig.PluginPath == "") {
+		return nil, ErrPlusNotEnabled
```
domenipavec, line 22: Not sure why error is returned if plus is not enabled, shouldn't the software just continue?

server/plus.go
```diff
@@ -0,0 +1,75 @@
+package chserver
```
domenipavec, line 1: Not sure why this file is in server, I think it should be part of the `plus` package.

plus/manager.go
```diff
@@ -0,0 +1,200 @@
...
+	} else {
+		// empty init func name indicates that the provider can be initialized locally
+		newCap.SetProvider(nil)
+	}
+
+	return newCap, err
+}
+
+// LoadInitFunc loads the capability init func symbol from the plugin
+func (pm *ManagerProvider) LoadInitFunc(pluginPath string, capName string) (sym plugin.Symbol, err error) {
```
domenipavec, line 117: Why is this method here, it seems it would be the same if you just call `LoadSymbol` directly?

plus/manager.go
```diff
@@ -0,0 +1,200 @@
...
+func (pm *ManagerProvider) GetCapability(capName string) (cap Capability) {
+	pm.mu.RLock()
+	defer pm.mu.RUnlock()
+
+	return pm.caps[capName]
+}
+
+// HasCapabilityEnabled returns whether the specified capability is present/enabled
+// or not
+func (pm *ManagerProvider) HasCapabilityEnabled(capName string) (isEnabled bool) {
```
domenipavec, line 140: This should be named either `HasCapability` or `IsCapabilityEnabled`.

rportd.example.conf
```diff
@@ -409,3 +409,69 @@
...
   ## Default: 30 days
   #data_storage_days = 30
+
+[plus-plugin]
+  ## Rport Plus is a paid for binary extension to Rport.
+  ## Learn more at https://plus.rport.io/
+  ## When using Rport Plus, the plugin path must be specified
+  # plugin_path = "./rport-plus.so"
+
+[plus-oauth]
```
domenipavec, line 419: This section of config is inconsistent. The comment describing the config value should be right above the commented out setting.

server/api_listener.go
```diff
@@ -393,7 +413,12 @@ func (al *APIListener) validateCredentials(username, password string, skipPasswo
 	if err != nil {
 		return false, nil, fmt.Errorf("failed to get user: %v", err)
 	}
-	if user == nil && skipPasswordValidation && al.config.API.CreateMissingUsers {
+
+	if user == nil && skipPasswordValidation &&
+		(al.config.API.CreateMissingUsers ||
+			// if we're using OAuth and there's no permitted_user_list then create
+			// any missing users
+			(al.config.PlusOAuthEnabled() && !al.config.OAuthConfig.PermittedUserList)) {
```
domenipavec, line 421: This if condition is getting a bit complicated, maybe wrap the check in a helper method like `if al.shouldCreateNewUser`.

server/api_router.go
```diff
@@ -177,6 +178,17 @@ func (al *APIListener) initRouter() {
...
 	})
 
+	plusRouter := api.PathPrefix(plusRoutesPrefix).Subrouter()
+	plusRouter.HandleFunc("/status", al.handlePlusStatus).Methods(http.MethodGet)
+
+	authRouter := api.PathPrefix(authRoutesPrefix).Subrouter()
+	authRouter.HandleFunc(authProviderRoute, al.handleGetAuthProvider).Methods(http.MethodGet)
+	authRouter.HandleFunc(authSettingsRoute, al.handleGetAuthSettings).Methods(http.MethodGet)
+
+	if al.config.PlusOAuthEnabled() {
```
domenipavec, line 188: I don't think we should conditionally setup routes, the handler should check if oauth is enabled and return error if it's not.

plus/capabilities/oauth/oauth.go
morestatic: hey domen, i'm open on how we do this. currently we're using map of capability interfaces to track the registered capabilities. we could move to a map of generic interfaces but overall i have a preference to maintain some type information. using the interface does also allow capability implementers a little more scope to setup on the rportd side of the interface. i think i'd prefer to stick with the capability interface and then see how we go. implementing the 3 fns of the interface isn't much work. shout if you feel strongly that isn't the right approach.

server/api_router.go
```diff
@@ -177,6 +178,17 @@ func (al *APIListener) initRouter() {
 		return nil
 	})
 
+	plusRouter := api.PathPrefix(plusRoutesPrefix).Subrouter()
```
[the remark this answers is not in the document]
morestatic: ok. i've removed the constant and replaced with the text. i know what you mean about being easier to read. there's a slight risk that other contributors will also use the literal text in other places which wouldn't be what we wanted, so we'll just need to keep an eye on that.

api-doc/openapi/openapi.yaml
```diff
@@ -29,9 +35,20 @@ tags:
...
   - name: Users
     description: For more details https://oss.rport.io/docs/no12-user.html
+  - name: Plus
+    description: |
+      For more details https://plus.rport.io/auth/oauth-introduction/
 paths:
   /login:
     $ref: paths/login.yaml
+  /auth/provider:
+    $ref: paths/auth-provider.yaml
```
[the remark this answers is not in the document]
morestatic: sometimes it seems to be underscore and sometimes dash. it isn't very consistent. should we change them all one way or the other? what do you think?

plus/manager.go
morestatic: they are currently only used in the tests but they are required if we're gonna maintain the tests in a _test package (rather than _test file). in general, i prefer using _test packages as they provide stronger decoupling from referenced packages, so there are less circular dependency errors. so it is a tradeoff between some additional public fns in interfaces vs worse coupling between packages. my preference is for the former as long as the goal is only to add such fns when it really doesn't make sense to only use the main interface fns. what do you think?

plus/manager.go
morestatic: moved the mutex up and moved the mutex locks / unlocks into `GetCapability` and `SetCapability`. agree that it is cleaner. see comment above regarding the exposure of these fns.

server/plus.go
morestatic: so for this, I agree in principle, but in implementation I ran into a bunch of circular dependency issues as the fns in `server/plus.go` reference both `chserver` definitions and `plus` package definitions. i could have just put the fns from the `plus` package into `chserver` but `chserver` feels too big already (in total approx 70 files), so i really wanted to avoid that. the current implementation is the compromise.

plus/manager.go
morestatic: agreed, it is somewhat of a wonky name. i've updated to `IsEnabledCapability`. not quite sure why I didn't go with `IsCapabilityEnabled` but I'd committed already by the time I was wondering why I was using the name I did, so just decided to leave it. let me know if you think we should change, otherwise I'll assume it is ok.

rportd.example.conf
morestatic: This is applied inconsistently in the example conf (see the `smtp` section) but I agree that notes just before the setting itself is better, especially when many settings, so i've updated as per your suggestion.

api-doc/openapi/openapi.yaml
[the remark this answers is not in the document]
domenipavec: Sometime there's dashes in urls, like for `client-groups`. Those should stay dashes, but all slashes should be underscores. If we're not consistent in that then we should fix it.

plus/manager.go
domenipavec: I don't have a strong preference for using `_test` package. It's purpose should be to test the package using only the public interface, not to have to expand the interface for tests. If you cannot test the package using only the public interface, then you shouldn't use `_test` package rather then change the interface.

cmd/rportd/main.go
```diff
@@ -460,7 +460,17 @@ func runMain(*cobra.Command, []string) {
 	// Flush the in-memory logger
 	mLog.Flush(logger.NewLogger("server-startup", cfg.Logging.LogOutput, cfg.Logging.LogLevel))
 
-	s, err := chserver.NewServer(cfg, files.NewFileSystem())
+	filesAPI := files.NewFileSystem()
+
+	plusManager, err := chserver.EnablePlusIfLicensed(cfg, filesAPI)
```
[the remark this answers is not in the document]
morestatic: the reason that not in NewServer is the the plugin enabling (via the capability registration) also validates the config and if we've separated out the config validation for the main server then I didn't want to have NewServer failing due to config validation (for plus). we could refactor to separate out the config validation for plus, but not sure we really get much value for doing so, for the additional complexity. also, i like the idea that NewServer is really the new server startup and that the plus plugin is all ready to go for that startup

plus/manager.go
morestatic: ok. i've expanded the mocks so we don't need the additional fns and i prefer the result, so i'm good with the change. however, i'm still concerned about the `chserver` package with 70 files (based on a very quick search). if a consequence of following the 'public interfaces only' principle tightly is that everything ends going into a single package (so that all interfaces/types are effectively open - to those 70 files) then i'm not sure we're making the right tradeoffs on our principles.

server/plus.go
morestatic: yes, we should be continuing. it is fixed in a later branch. not quite sure how we'll do the feedback resolution. i'll probably need to include links to revisions in the later branches here in the comments.

plus/manager.go
morestatic: yes, i think that's better, so i've removed LoadInitFunc

api-doc/openapi/paths/plus-status.yaml
```diff
@@ -0,0 +1,31 @@
...
+    "200":
+      description: Success
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              data:
+                $ref: ../components/schemas/PlusStatusResponse.yaml
+    "403":
```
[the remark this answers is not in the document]
morestatic: i'm not sure. `Forbidden` seems like a valid response. are you thinking that `Forbidden` will cause clients some confusion? can you elaborate a bit?

docs/.markdownlint.yaml
```diff
@@ -20,4 +20,7 @@ MD013:
   # Strict length checking
   strict: false
   # Stern length checking
-  stern: false
\ No newline at end of file
+  stern: false
+
+# TODO: disabled due to bug in markdown lint (i think). can be re-enabled in the future.
```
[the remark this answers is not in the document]
morestatic: the issue was related to bullets in MD. However, either they've fixed the issue or we're not using bullets in the same way now. i removed the disabled check and the lint is no longer failing. i've updated the markdown lint yaml file.

api-doc/openapi/openapi.yaml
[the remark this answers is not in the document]
morestatic: ok. i've updated the plus related filenames but will leave the others. i'm not really clear when something is a url or not.

server/api_listener.go
morestatic: ok. yes agreed. moved checks to their own fn.

server/config.go
morestatic: i quite liked having the capability configs here. not sure i quite understand your goal? the oauth and plugin configs are separated. i could understand making the capability config less visible to rport. i think only one of the oauth cap config fields are used in rport, so arguably the full cap config wouldn't need to have the current visibility. but i'm not sure that's what you're thinking?

server/api_router.go
morestatic: i'm not sure on this one. i agree in principle but with something like Plus it isn't clear to me that there's a potential benefit from exposing all the endpoints when the functionality just isn't there/available. what are you thinking the consequences (in this scenario) of not exposing the routes will be?

api-doc/openapi/paths/plus-status.yaml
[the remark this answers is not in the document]
domenipavec: Yes, I think `Forbidden` is usually related to authentication, but this is more related to configuration. I would bring it up with Thorsten, but if he thinks `Forbidden` is the way to go, then that's fine with me.

server/config.go
domenipavec: I was thinking like this:

`server/config.go`:
```
PlusConfig *rportplus.PlusConfig `mapstructure:",squash"`
```

`plus/config.go`:
```
type PlusConfig struct {
	PluginConfig *PluginConfig `mapstructure:"plus-plugin"`
	OAuthConfig *oauth.Config `mapstructure:"plus-oauth"`
}
type PluginConfig struct {
	PluginPath string `mapstructure:"plugin_path"`
}
```

This way only top level `PlusConfig` is needed in server.

cmd/rportd/main.go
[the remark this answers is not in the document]
domenipavec: ok

plus/manager.go
domenipavec: The reason `chserver` is huge is that all the code for api handlers is tightly coupled and hard to separate, but we should definitely split the package at some point. I don't think having tests in the same package as the code really has much effect on this.

server/api_router.go
domenipavec: I don't see a huge downside of doing this, I just don't think it's necessary. It might also make it harder to understand the code if something causes some endpoints to disappear.
But feel free to leave this as is if you disagree.

server/config.go
morestatic: ok. done.

server/api_router.go
morestatic: ok. i think i will leave it, on the basis that if i was running OSS rport i wouldn't want rport-plus endpoints responding (even if only returning a disabled response).

domenipavec (approved):

PR #714, opened 2022-10-10
[DEV-2666, DEV-2667] Add Google as provider, Implement OAuth for cli

morestatic:
This PR includes functionality added and PR feedback updates since [PR #699](https://github.com/cloudradar-monitoring/rport/pull/699) ([DEV-2618, DEV-2617, DEV-2643] Prepare rport-plus sso / oauth for release) was raised. The PR merges these changes into that PR. Once [PR #699](https://github.com/cloudradar-monitoring/rport/pull/699) has been accepted then all changes will be merged into main.

See the commit history for a detailed list of changes.


server/api_handler_plus.go
```diff
@@ -47,6 +47,56 @@ func (al *APIListener) handleOAuthAuthorizationCode(w http.ResponseWriter, r *ht
...
+		if errInfo != nil {
+			// error handling for the OAuth device flow is a little strange.
+			// pending and slow_down errors aren't really errors. Also
+			// the different providers seems to sometimes report errors via
+			// the statusCode and sometimes not. Do the best that we can
+			// here.
+			response := api.NewSuccessPayload(errInfo)
+			al.writeJSONResponse(w, errInfo.StatusCode, response)
+		} else {
+			// TODO: i think we need a better status code for the response
```
domenipavec, line 79: Either change the code to `http.StatusInternalServerError` or leave it, but please remove the TODO.

server/api_handler_plus_test.go
```diff
@@ -305,6 +402,17 @@ func SetupAPIListener(t *testing.T, oauthConfig *oauth.Config, username string)
 	})
 	require.NoError(t, err)
 
+	al, mockUsersService = MakeAPIListener(t, plusManager, plusConfig, oauthConfig)
+
+	return al, mockUsersService
+}
+
+func MakeAPIListener(
```
domenipavec, line 410: This function name is too generic for the server package. Please make it more specific for this file. Otherwise it might end up used all over the server package and it'll make it even harder to potentially split the package in the future.

server/api_handler_plus_test.go
```diff
@@ -285,6 +325,63 @@ func TestHandleOAuthAuthorizationCode(t *testing.T) {
...
+	req := httptest.NewRequest("GET", "/api/v1"+oauth.DefaultDeviceLoginURI, nil)
+
+	al.router.ServeHTTP(w, req)
+
+	assert.Equal(t, http.StatusOK, w.Code)
+
+	assert.Contains(t, w.Body.String(), `{"data":{"token":"`)
+}
+
+type AuthStatusErrorResponse struct {
```
domenipavec, line 349: Nit.: I'm not sure I like a separate struct definition just for decoding responses in tests. I know we don't have a good way of doing this currently. Maybe this would be a good time to start using generics and refactor `SuccessPayload` to use generics so it could be used for decoding as well?

domenipavec (commented):
In the future please make the changes that address PR comments in the same PR, otherwise it makes it harder to review then is necessary.
Each PR should be a self contained change, preferably as small as possible.

server/api_handler_plus_test.go
morestatic: ok. renamed.

server/api_handler_plus_test.go
morestatic: yes agreed. done in a limited form with generics. if we like the approach then we can take further. added in a new file called `api_login_auth_provider.go`.

server/api_handler_plus.go
morestatic: done.

plus/capabilities/oauth/oauth.go
```diff
@@ -28,12 +30,61 @@ type LoginInfo struct {
 	Expiry       time.Time `json:"expiry"`
 }
 
-// CapabilityEx represents the functional interface provided by the OAuth capability
+// DeviceAuthInfo contains the info returned when getting auth settings
+// for the device style flow
+type DeviceAuthInfo struct {
```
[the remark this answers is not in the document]
morestatic: ok. the file is mostly the definitions, but what i've done is move the capability stub code into a new file called `oauthcap.go`. then at least the two concerns are now in separate files, which is a little cleaner.

cmd/rportd/main_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"testing"
+
+	"github.com/cloudradar-monitoring/rport/share/logger"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestShouldLoadConfigFromFile(t *testing.T) {
```
domenipavec, line 11: We already have the test in `config_load_test.go` that tries to load the example config. You should probably add your checks there or combine the tests.

Also not sure if `test.conf` contains some differences from example config, but if possible you should test with example config.
If that's not possible, then I think you should refactor the loading code to take a reader, so you can load config from strings in test code. Also those should be minimal config snippets required for tests, not full blown configs.

server/api_login_auth_provider_test.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// SuccessPayloadResponseParam provides type constraints for using the SuccessPayloadResponse
+// generic type as a fn param
+type SuccessPayloadResponseParam interface {
+	SuccessPayloadResponse[AuthProviderInfo] | SuccessPayloadResponse[oauth.DeviceAuthStatusErrorInfo]
+}
+
+// GetSuccessPayloadResponse returns a successful payload response of the expected type
+func GetSuccessPayloadResponse[R SuccessPayloadResponseParam](r io.Reader, response *R) (err error) {
```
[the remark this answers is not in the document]
morestatic: good suggestion. done.

cmd/rportd/main_test.go
morestatic: ok. i missed that test. not sure why cause it is sort of obvious. it is little tricky testing against the example.conf because we don't want to enable rport-plus in the example config. will think about it but will get rid of the `main_test.go` file and move any new test to the `config_load_test.go` file.

morestatic: I've added a new test in `config_load_test.go`. it generates a small conf file with the plus params and test loads that, then removes it after. not sure whether the best approach but felt better than keep a specific test file in the repo.

morestatic: So i've committed a new test that just generates (and removes) a temporary small config file. Only noticed your comment about the reader afterwards (i.e. just now). Reader would be cleaner but involves the refactoring, so I'll take a look at that in the morning.

plus/capabilities/oauth/oauth.go
```diff
@@ -49,51 +107,19 @@ type Config struct {
...
-// GetOAuthCapabilityEx returns the interface to the capability functions
-func (cap *Capability) GetOAuthCapabilityEx() (capEx CapabilityEx) {
-	return cap.Provider
-}
 
-// GetConfigValidator returns a validator interface that can be called to
-// validate the capability config
-func (cap *Capability) GetConfigValidator() (v validator.Validator) {
-	return cap.Provider
+	CompiledPermittedUserMatch *regexp.Regexp
```
domenipavec, line 124: This is not actually used anywhere?

domenipavec: Never mind, I see this is needed from plus code.

domenipavec (approved):

PR #790, opened 2022-11-15
Feature/dev 2684 password invalidation

edoardoc:
https://tracker.rport.io/issue/DEV-2684


server/api_listener.go
```diff
@@ -363,6 +364,10 @@ func (al *APIListener) handleBasicAuth(username, password string) (authorized bo
 		return false, username, nil
 	}
 
+	if user.PasswordExpired {
```
domenipavec, line 367: This function also handles token authentication as well as password. I think in case the token is used, this check should not happen.

server/api/users/user.go
```diff
@@ -2,12 +2,14 @@ package users
...
-	Username    string   `json:"username" db:"username"`
-	Password    string   `json:"password" db:"password"`
-	Groups      []string `json:"groups" db:"-"`
-	TwoFASendTo string   `json:"two_fa_send_to" db:"two_fa_send_to"`
-	Token       *string  `json:"token,omitempty" db:"token"`
-	TotP        string   `json:"totp_secret,omitempty" db:"totp_secret"`
+	Username              string   `json:"username" db:"username"`
+	Password              string   `json:"password" db:"password"`
+	PasswordExpired       bool     `json:"-" db:"password_expired"`
+	ActionPasswordExpired string   `json:"password_expired" db:"-"`
```
domenipavec, line 8: Why is this a separate field? And why is it a string?

server/api_handler_login.go
```diff
@@ -82,6 +83,24 @@ func (al *APIListener) handleLogin(username, pwd string, skipPasswordValidation
...
+		var user users.User
+		user.Password = newpwd
+		if err := al.userService.Change(&user, username); err != nil {
+			al.jsonError(w, err)
+			return
+		}
+		return
+	}
+
+	// is it expired?
```
domenipavec, line 98: This comment is basically the same as code.

server/api_handler_login.go
```diff
@@ -188,7 +207,9 @@ func (al *APIListener) handlePostLogin(w http.ResponseWriter, req *http.Request)
 		return
 	}
 
-	username, pwd, err := parseLoginPostRequestBody(req)
+	// you can let through only the newPassword field from here (not from GET)
```
domenipavec, line 210: I'm not sure I understand this comment.

server/api_handler_login.go
```diff
@@ -82,6 +83,24 @@ func (al *APIListener) handleLogin(username, pwd string, skipPasswordValidation
 		return
 	}
 
+	// password is correct,
+	// a new password is provided?
```
domenipavec, line 87: Change this to single line comment. I would also make it a sentence, like:
`Only set the new password after the old password has been verified.`

domenipavec (changes requested):
The ticket mentions implementations for both db and json user stores. Are you going to do the json implementation in separate PR?

server/api/users/database.go
```diff
@@ -356,6 +361,16 @@ func (d *UserDatabase) Update(usr *User, usernameToUpdate string) error {
 			return err
 		}
 	}
+	if usr.Username == "" && usr.Password != "" { // I am updating the password
```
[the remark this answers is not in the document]
edoardoc: cause its a specific case, where password_expired goes to false at the same time as a new passwors gets in.

(similar to the case below where the username gets updated)

server/api/users/database.go
```diff
@@ -51,7 +51,7 @@ func NewUserDatabase(
 }
 
 func (d *UserDatabase) getSelectClause() string {
-	s := "username, password"
+	s := "username, password, password_expired"
```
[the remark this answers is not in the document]
edoardoc: no, I could not see the compatibility-with-old-db-versions in the task, indeed this would not work in that case.
For your second statement, checkDatabaseTables checks all the columns via getSelectClause and by calling it in a select; so it will throw an error when not present.

server/api_listener.go
edoardoc: per @thorstenkramm instructions I was told "The user must be kicked out immediately" and for the token I am to await another job to finish so to invalidate the token also (possibly in the same context, I am not sure)

server/api/users/user.go
edoardoc: Since the field itself is a boolean, but no update can be decided upon a boolean only field (cause you have no "presence" check),
I decided to add an API only string field ActionPasswordExpired instead of some added logic, but if you have an easier way we can try and implement that.

server/api_handler_login.go
edoardoc: leftover, removed

server/api_handler_login.go
edoardoc: done

server/api_handler_login.go
edoardoc: ok, will try and make it more understandable

server/api_listener.go
domenipavec: I understand the user session must be stopped. But user tokens are not the same as jwt tokens. Jwt tokens are used for user sessions after the initial password login. But user tokens are long lived tokens that are meant to be used in scripts.
I don't think those should stop working on password invalidate. But if that's @thorstenkramm 's requirement, that'sFine with me.

server/api/users/database.go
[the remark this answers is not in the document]
domenipavec: I'm not sure it needs to be a specific case, what's stopping it from being updated simultaneously?

server/api/users/user.go
```diff
@@ -27,6 +28,10 @@ func (u User) IsAdmin() bool {
 	return false
 }
 
+func PasswordExpired(f bool) *bool {
```
domenipavec, line 31: We have `ptr.Bool` in `share` that could be used.

server/api/users/database.go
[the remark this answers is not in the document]
domenipavec: Ok, as long as we note that everyone will need to do a database migration when upgrading to the new version.

domenipavec (approved):

server/api/users/database.go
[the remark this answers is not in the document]
edoardoc: yes, this block is useless, I have removed it. Thanks

server/api/users/user.go
edoardoc: I checked the rest of the repo, it is seldom used.

edoardoc:
> The ticket mentions implementations for both db and json user stores. Are you going to do the json implementation in separate PR?

I have done that also

domenipavec (approved):

server/api/users/database.go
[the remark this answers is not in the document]
edoardoc: I have added notes in the db creation docs

domenipavec (approved):

PR #792, opened 2022-11-15
Fixes for update checks on CentOS-like Redhat OS using yum

jackyaz:
These changes allow rport to successfully list and detail updates available on CentOS.

Tested on CentOS 7, testing on later CentOS/Rocky/Alma as well as RHEL is advised.

Notes from comments:

Security updates will always be 0 on CentOS-like Redhat without EPEL, as core CentOS doesn't include security metadata
"needs-restarting" requires the yum-utils package

### DIFF
--- a/client/updates/yum.go
+++ b/client/updates/yum.go
@@ -38,7 +38,7 @@ func (p *YumPackageManager) GetUpdatesStatus(ctx context.Context, logger *logger
 	// --refresh doesn't work on CentOS, replace with clean expire-cache which should be supported by both dnf and yum
 	cmd := append([]string{"clean", "expire-cache", "--quiet"})
         _, err := p.run(ctx, cmd...)
-	if err != nil && err.Error() != "exit status 1" {
+	if err != nil {
         	return nil, err
         }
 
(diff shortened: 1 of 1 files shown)


client/updates/yum.go
```diff
@@ -34,11 +35,19 @@ func (p *YumPackageManager) IsAvailable(ctx context.Context) bool {
 }
 
 func (p *YumPackageManager) GetUpdatesStatus(ctx context.Context, logger *logger.Logger) (*models.UpdatesStatus, error) {
-	allUpdates, err := p.listUpdates(ctx, "--refresh")
+	// --refresh doesn't work on CentOS, replace with clean expire-cache which should be supported by both dnf and yum
+	cmd := append([]string{"clean", "expire-cache", "--quiet"})
+        _, err := p.run(ctx, cmd...)
+	if err != nil && err.Error() != "exit status 1" {
```
domenipavec, line 41: Please add a comment explaining when `exit status 1` happens and why it's ok to continue.

jackyaz: On reflection, I can't find a reason why this would be needed, so removed the test for exit code 1

domenipavec (approved):

PR #795, opened 2022-11-18
Move server config to own package

morestatic:
The current `chserver` package currently consists of 65 files. This feels too many and makes testing harder. It would be good to start to separate out some of the functionality into separate packages. Most likely doing this for some code will be easier than for other code and also separating out some code will result in higher value (in terms of future reduced cost of change, etc.) compared to other code. We should focus on low cost and high value refactorings initially.

This PR separates out the server config data structures and functionality into a separate `chconfig` package. The refactoring was fairly straightforward. When adding new functionality there is quite often a dependency on the server config. This refactoring will allow the server config to be used without pulling in the `chserver` package.

Note: happy to rename the `chconfig` package to something else if we feel there's a better name.


domenipavec (approved):
I'm not sure what the best name is, but I would just keep it simple as `config`.

thorstenkramm (approved):

PR #803, opened 2022-11-22
[DEV-2700] Too many db locks

morestatic:
This PR adds explicit retry attempts when sqlite returns busy db errors (aka locked database errors). 

Other options considered were to use busy timeouts or restrict max open connections to 1. Using busy timeouts resulted in more busy errors (surprisingly) and limiting the max open connections lowered performance of writes to approx 1/2 the previous performance (not surprisingly).

https://tracker.rport.io/issue/DEV-2700/Too-many-db-locks

### DIFF
--- a/db/sqlite/sqlite.go
+++ b/db/sqlite/sqlite.go
@@ -67,8 +67,8 @@ func New(dataSourceName string, assetNames []string, asset func(name string) ([]
 	return db, nil
 }
 
-func WithRetryWhenBusy[R any](retryAble func() (result R, err error), label string, l *logger.Logger, maxAttempts int) (result R, err error) {
-	for r := 0; r < maxAttempts; r++ {
+func WithRetryWhenBusy[R any](retryAble func() (result R, err error), label string, l *logger.Logger) (result R, err error) {
+	for r := 0; r < DefaultMaxAttempts; r++ {
 		result, err = retryAble()
 		if err != nil {
 			sqlErr, ok := err.(sql.Error)
--- a/db/sqlite/sqlite.go
+++ b/db/sqlite/sqlite.go
@@ -3,14 +3,22 @@ package sqlite
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/golang-migrate/migrate/v4"
 	"github.com/golang-migrate/migrate/v4/database/sqlite3"
 	bindata "github.com/golang-migrate/migrate/v4/source/go_bindata"
 	"github.com/jmoiron/sqlx"
+	sql "github.com/mattn/go-sqlite3"
+
+	"github.com/cloudradar-monitoring/rport/share/logger"
 )
 
-const WALEnabled = "_journal_mode=WAL"
+const (
+	WALEnabled                  = "_journal_mode=WAL"
+	defaultDelayBetweenAttempts = 10 * time.Millisecond
+	DefaultMaxAttempts          = 3
+)
 
 type DataSourceOptions struct {
 	WALEnabled bool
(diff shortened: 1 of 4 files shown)


db/sqlite/sqlite.go
```diff
@@ -3,14 +3,22 @@ package sqlite
...
 	"github.com/jmoiron/sqlx"
+	sql "github.com/mattn/go-sqlite3"
+
+	"github.com/cloudradar-monitoring/rport/share/logger"
 )
 
-const WALEnabled = "_journal_mode=WAL"
+const (
+	WALEnabled                  = "_journal_mode=WAL"
+	defaultDelayBetweenAttempts = 10 * time.Millisecond
```
domenipavec, line 19: Is 10ms chosen at random or is it the optimal value for maximum write performance?

db/sqlite/sqlite.go
```diff
@@ -58,3 +66,24 @@ func New(dataSourceName string, assetNames []string, asset func(name string) ([]
...
 }
+
+func WithRetryWhenBusy[R any](retryAble func() (result R, err error), label string, l *logger.Logger, maxAttempts int) (result R, err error) {
+	for r := 0; r < maxAttempts; r++ {
+		result, err = retryAble()
+		if err != nil {
+			sqlErr, ok := err.(sql.Error)
+			if ok && sqlErr.Code == sql.ErrBusy {
+				l.Debugf("%s: attempt %d: source err = %+v\n", label, r+1, err)
+				time.Sleep(defaultDelayBetweenAttempts)
```
domenipavec, line 77: Not sure just retrying at fixed intervals is optimal. Common practice is to use exponentially larger delays and add some randomness.

db/sqlite/sqlite.go
```diff
@@ -58,3 +66,24 @@ func New(dataSourceName string, assetNames []string, asset func(name string) ([]
...
 	return db, nil
 }
+
+func WithRetryWhenBusy[R any](retryAble func() (result R, err error), label string, l *logger.Logger, maxAttempts int) (result R, err error) {
+	for r := 0; r < maxAttempts; r++ {
+		result, err = retryAble()
+		if err != nil {
+			sqlErr, ok := err.(sql.Error)
+			if ok && sqlErr.Code == sql.ErrBusy {
+				l.Debugf("%s: attempt %d: source err = %+v\n", label, r+1, err)
```
domenipavec, line 76: Label is a bad way to do this. The caller should just fork the logger with appropriate prefix instead of having this label here.

db/sqlite/sqlite.go
```diff
@@ -58,3 +66,24 @@ func New(dataSourceName string, assetNames []string, asset func(name string) ([]
 
 	return db, nil
 }
+
+func WithRetryWhenBusy[R any](retryAble func() (result R, err error), label string, l *logger.Logger, maxAttempts int) (result R, err error) {
```
domenipavec, line 70: I don't think it's worth having `maxAttempts` as an argument if it's always the same.

db/sqlite/sqlite.go
morestatic: it is pretty random really but not completely unconsidered. it isn't too long to be a real drag on the write but probably not too short that a previous write is unlikely to have completed. without setting up a performance test environment probably tricky to determine an optimal value and even then depends on the customer environment itself. naturally with the this approach we'll need to see if any negative feedback from the field. one mitigation in the future would be to make the value customer configurable.

db/sqlite/sqlite.go
morestatic: i think a label is the right way to go in this scenario. the log is really in the context of the caller and lives for just that call, so should use the that logger and not a fork. the label is just to clarify which fn the closure belongs to. agree that in other scenarios, fork is a much better approach.

db/sqlite/sqlite.go
morestatic: ok. yes agree that we'll probably not change `maxAttempts` once we're happy with the number. i'll remove it.

morestatic: done.

domenipavec (approved):
It would be nice to have a go benchmark for this so we could determine a good value for delay and see if adding some randomness would help distribute the load of retries.

PR #808, opened 2022-11-24
[DEV-2709] Improve logging

morestatic:
This PR improves aspects of our general logging approach. We have a customer who's having problems with disconnected clients and connected tunnels being out of sync. It is proving difficult to diagnose as our logs are missing certain information about client disconnects / connects. This PR adds some of the missing info.

The PR also contains a change that isn't specifically related to improved logging that might assist with the customer problem where we will attempt to start tunnels even if the client is marked as disconnected. Ideally this would be a separate PR but in the interests of deploying a change quickly, it has been included with PR.

Also, there is a change, that will be enhanced further after the next pre-release, to log api error responses on the server side. We currently have a number of internal server error responses that aren't logged on the server. A future change will move these logs out from the server log into a separate (and configurable) api responses log.

### DIFF
--- a/server/api_handler_clients.go
+++ b/server/api_handler_clients.go
@@ -336,9 +336,23 @@ func (al *APIListener) handlePutClientTunnel(w http.ResponseWriter, req *http.Re
 		al.jsonErrorResponse(w, http.StatusInternalServerError, err)
 		return
 	}
+
+	attemptingClientReconnect := false
 	if client == nil {
-		al.jsonErrorResponseWithTitle(w, http.StatusNotFound, fmt.Sprintf("client with id %s not found", clientID))
-		return
+		al.Debugf("active client with id %s not found", clientID)
+		// there isn't an active client, but let's see if there's a disconnect client and if so then
+		// try creating tunnels anyway by continuing with the non-active client.
+		attemptingClientReconnect = true
+		client, err = al.clientService.GetByID(clientID)
+		if err != nil {
+			al.jsonErrorResponse(w, http.StatusInternalServerError, err)
+			return
+		}
+		if client == nil {
+			// still not found
+			al.jsonErrorResponseWithTitle(w, http.StatusNotFound, fmt.Sprintf("client with id %s not found", clientID))
+			return
+		}
 	}
 
 	localAddr := req.URL.Query().Get("local")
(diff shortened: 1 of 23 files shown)


server/api_handler_clients.go
```diff
@@ -336,9 +336,23 @@ func (al *APIListener) handlePutClientTunnel(w http.ResponseWriter, req *http.Re
...
 		return
 	}
+
+	attemptingClientReconnect := false
 	if client == nil {
-		al.jsonErrorResponseWithTitle(w, http.StatusNotFound, fmt.Sprintf("client with id %s not found", clientID))
-		return
+		al.Debugf("active client with id %s not found", clientID)
+		// there isn't an active client, but let's see if there's a disconnect client and if so then
+		// try creating tunnels anyway by continuing with the non-active client.
```
domenipavec, line 344: This seems like a bad idea and will probably lead to even more out of sync clients and tunnels if we allow to create active tunnels that don't have a backing connected client.

server/api_responses.go
```diff
@@ -4,11 +4,28 @@ import (
...
+	"os"
 
 	"github.com/cloudradar-monitoring/rport/server/api"
 	errors2 "github.com/cloudradar-monitoring/rport/server/api/errors"
+	"github.com/cloudradar-monitoring/rport/share/logger"
 )
 
+var (
+	// this will be used by default for tests, but otherwise should be overridden during server init
+	errLog = logger.NewLogger("api-error-response", logger.LogOutput{File: os.Stdout}, logger.LogLevelDebug)
```
domenipavec, line 16: Is there a reason to use a global variable here? I think it should be part of APIListener.

server/client_service.go
```diff
@@ -409,6 +415,12 @@ func (s *ClientServiceProvider) startClientTunnels(client *clients.Client, remot
 		}
 		tunnels = append(tunnels, t)
 	}
+
+	// if we created the tunnels with a disconnected client then force reconnect the client
```
domenipavec, line 419: This will not actually reconnect the client, it will just make it appear to be connected. This comment is misleading. Also the state of the client connection should be easily determined - either we have an active connection or we don't. Forcing the state without an active connection seems useless to me.

domenipavec (changes requested):

server/api_handler_clients.go
thorstenkramm: @morestatic @domenipavec 
The intention of this change is to determine if the client status stored in memory is correct.
We want this just for testing purposes. Will be reset to the previous implementation on the next release. 
We have user feedback about clients whose status changed to disconnected while a tunnel (SSH connection) was active and usable. This suggests the assumption that the client was not disconnected, just the information stored on the server, was false. We want to test what happens if we create a tunnel to a client that is shown disconnected on the server but connected on the client.

server/api_responses.go
morestatic: moved API Listener. done.

server/client_service.go
morestatic: removed this. done.

thorstenkramm (approved):

PR #811, opened 2022-11-28
Feature/dev 2685 password strength

edoardoc:
https://tracker.rport.io/issue/DEV-2685/Password-strength


rportd.example.conf
```diff
@@ -357,6 +357,20 @@
   ## Consider changing to a faster rotation.
   #audit_log_rotation = 'monthly', possible values: yearly, monthly, weekly, daily
 
+  ## Required minimal password length
+  ## Default: 14
+  # password_min_length = 14
```
domenipavec, line 362: There's usually no space after `#` for example configs in this file.

server/api/users/service.go
```diff
@@ -208,6 +211,24 @@ func (as *APIService) validate(dataToChange *User, usernameToFind string) error
...
 
+	if dataToChange.Password != "" {
+		if len(dataToChange.Password) < as.PasswordMinLength {
+			errs = append(errs, errors2.APIError{
+				Message:    fmt.Sprintf("password must be at least %v characters", as.PasswordMinLength),
+				HTTPStatus: http.StatusBadRequest,
+			})
+		}
+		if as.PasswordZxcvbnMinscore >= 0 { // -1 means no zxcvbn
+			score := zxcvbn.PasswordStrength(dataToChange.Password, nil)
```
domenipavec, line 222: I think we should pass username as `userInputs` here.

rportd.example.conf
```diff
@@ -357,6 +357,20 @@
...
+  
+  ## Perform the so-called zxcvbn check to reject passwords based on common words.
+  ## Enabled by default. The zxcvbn check is executed locally. No API calls.
+  ## Default 0 (minimum risky password: "too guessable")
+  ## 1  modest protection from throttled online attacks: "very guessable"
```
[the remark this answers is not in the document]
edoardoc: done

rportd.example.conf
edoardoc: done

server/api/users/service.go
```diff
@@ -144,6 +149,10 @@ func (as *APIService) ExistGroups(groups []string) error {
 }
 
 func (as *APIService) Change(usr *User, username string) error {
+	err := as.validate(usr, username) //validate does it all validation
```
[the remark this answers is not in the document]
edoardoc: my intent was to remind anybody that validation logic should not go elsewhere, maybe we can write something else? anyway, removing it for now

server/api/users/service.go
edoardoc: I don't understand

domenipavec: Second argument to `PasswordStrength` is `userInputs` slice of strings. You should pass in any other user properties like username. They're used to compare against the password, so it cannot be similar to them.

edoardoc: done

server/api/users/service.go
```diff
@@ -208,6 +211,24 @@ func (as *APIService) validate(dataToChange *User, usernameToFind string) error
...
 
+	if dataToChange.Password != "" {
+		if len(dataToChange.Password) < as.PasswordMinLength {
+			errs = append(errs, errors2.APIError{
+				Message:    fmt.Sprintf("password must be at least %v characters", as.PasswordMinLength),
+				HTTPStatus: http.StatusBadRequest,
+			})
+		}
+		if as.PasswordZxcvbnMinscore >= 0 { // -1 means no zxcvbn
+			score := zxcvbn.PasswordStrength(dataToChange.Password, []string{usernameToFind})
```
domenipavec, line 222: When creating a new user `usernameToFind` will be empty. I'm not sure how `PasswordStrength` behaves if you pass it empty string.
In any case it would be better to use `dataToChange.Username` when it's not empty and `usernameToFind` otherwise.

domenipavec: Or you could do something like:
```
var userInputs []string
if usernameToFind != "" {
  userInputs = append(userInputs, usernameToFind)
}
if dataToChange.Username != "" {
  userInputs = append(userInputs, dataToChange.Username)
}
```

edoardoc: yes, it makes sense: in this way - also - the user creation endpoint will inhibit passwords that contain that username.
Adding the new section... thanks

domenipavec (approved):

PR #821, opened 2022-12-02
[DEV-1934] Refactor code related to client sessions

morestatic:
This PR mostly moves the client service functionality into it's own package. Part of the ongoing work to separate out some of the concerns of the chserver package to reduce implicit coupling and improve test-ability etc.

Also included is a change to revert the creation of tunnels even when a non-active client.


server/clientservice/client_service.go
```diff
@@ -1,4 +1,4 @@
-package chserver
+package clientservice
```
domenipavec, line 1: I'm not sure I like this package living in parallel with `clients` package, when it mostly wraps that package functionality. I would either merge it with `clients` package or move what's currently in the `clients` to `clients/repository` and move this to `clients`.

morestatic: don't disagree with this. i've moved the client services functionality into the clients package. commit pushed.

domenipavec (approved):

PR #823, opened 2022-12-02
[DEV-2687] Use encoding for writing script file when available

domenipavec:
https://tracker.rport.io/issue/DEV-2687/broken-support-for-special-characters-on-windows

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -52,6 +52,7 @@ type Client struct {
 	monitor            *monitoring.Monitor
 	serverCapabilities *models.Capabilities
 	consoleDecoder     *encoding.Decoder
+	consoleEncoder     *encoding.Encoder
 	filesAPI           files.FileAPI
 	watchdog           *Watchdog
 }
@@ -103,6 +104,7 @@ func NewClient(config *ClientConfigHolder, filesAPI files.FileAPI) (*Client, err
 	if enc != nil {
 		logger.Infof("Console encoding detected as: %s", enc)
 		client.consoleDecoder = enc.NewDecoder()
+		client.consoleEncoder = enc.NewEncoder()
 	}
 
 	return client, nil
--- a/client/client.go
+++ b/client/client.go
@@ -21,7 +21,6 @@ import (
 	"github.com/shirou/gopsutil/host"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/net/proxy"
-	"golang.org/x/text/encoding"
 
 	"github.com/cloudradar-monitoring/rport/client/monitoring"
 	"github.com/cloudradar-monitoring/rport/client/system"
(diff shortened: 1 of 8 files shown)


thorstenkramm:
> I handled this mess in rex-powershell by [b64 encoding of the unicode codepoints](https://github.com/rapid7/rex-powershell/blob/master/lib/rex/powershell/output.rb#L79) of the input. You guys are more than welcome to pillage that repo - been used to successfully pwn targets in various forms for >10y 😄

@sempervictus If I understand the approach inside the ruby file right, it takes the script output after the script has finished and wraps it into a compress base64 encoding. This would break our script output streaming. We don't wait for the script to finish. Output is streamed to the server almost real time. We would have to use the approach for each line of output. But this seems like an overhead to me. 

We keep trying to convert the encoding with the golang built-in methods first. Anyway, thanks for the hint. We keep this approach in mind.

domenipavec:
@thorstenkramm I tested this with cmd, PowerShell and bash.

I'll add support for configuring custom encoding for interpreter aliases like we discussed in a separate PR.

sempervictus:
> > I handled this mess in rex-powershell by [b64 encoding of the unicode codepoints](https://github.com/rapid7/rex-powershell/blob/master/lib/rex/powershell/output.rb#L79) of the input. You guys are more than welcome to pillage that repo - been used to successfully pwn targets in various forms for >10y smile
> 
> @sempervictus If I understand the approach inside the ruby file right, it takes the script output after the script has finished and wraps it into a compress base64 encoding. This would break our script output streaming. We don't wait for the script to finish. Output is streamed to the server almost real time. We would have to use the approach for each line of output. But this seems like an overhead to me.
> 
> We keep trying to convert the encoding with the golang built-in methods first. Anyway, thanks for the hint. We keep this approach in mind.

You've got the input part right for sure (that unicode conversion piece is key before b64), but the output streams exactly the same way as invoking a script file or `powershell -c "..."`. For encoded commands over 8k, there is a chunking approach which takes every 8k chars, assigns them to a shell var, and then `powershell -e %VAR1%+%VAR2%...` to bypass the `cmd` input length restriction. There is also a compression/decompression harness which we use to reduce the size of the initial PSH script prior to encoding passes. As i said, you're free to pillage that lib - i'm the author of most of it (the git history is mangled, we extracted the Rex gems from Metasploit-proper a few years ago and broke our history but good, you can find the bulk of the effort in Msf's own git history).

If your intent is to run interactive holding IOs to the PID running PSH, then i think you're left with performing the `toByte`/`toEncodedChar` type conversions at either end of the IO stream, which i would probably do in the PSH side to ensure that the IOs are only handling raw bytes (not dealing with encoding at all). The difficulty with interactive REPL over remote sessions is that PSH uses XML under the skin for its representation and errors thrown in that format do not translate well to line-wise readers (i never dug in deep enough to find why the outputs get trunc'd like they do). Output-wise, there's also a bunch of places running non-unicode charsets, and many of them are in financial conditions where free tooling is the only tooling they can afford - so a charset detector and relevant conversion selector on the PSH side might provide a sort of bablefish to avoid having to deal with any of that in Go.

thorstenkramm (approved):

PR #828, opened 2022-12-06
Fix/dev 2723 windows client process dead but running

thorstenkramm:
I've seen logs where the client seems to get stuck on the initial connection. Because of the missing time out, the client never retries. 
Furthermore, I'm wondering why a connection error on `client/client.go:376` was not considered a retriable error. On errors, the client was terminating there. I think there is no reason why a client should stop retrying.


domenipavec (approved):
Yes, this should be retriable, was probably just missed.

PR #830, opened 2022-12-07
Fix/dev 2725 tunnel db saving

thorstenkramm:
@morestatic Your observation was right. Tunnel removal was not saved to the db. Deleted tunnels are wrongly coming up again after server restart. 
I don't know if it's a good or silly idea to add the client repository to the client. But it gives a convenient way to persist the client from wherever the client is changed. Be critical!

I don't understand why the two disabled tests are now failing with a nil pointer dereference. I need your help. 🥺

### DIFF
--- a/server/api_handler_clients.go
+++ b/server/api_handler_clients.go
@@ -412,7 +412,7 @@ func (al *APIListener) handlePutClientTunnel(w http.ResponseWriter, req *http.Re
 		remote.Scheme = &schemeStr
 	}
 
-	if existing := client.FindTunnelByRemote(remote); existing != nil {
+	if existing := al.clientService.FindTunnelByRemote(client, remote); existing != nil {
 		al.jsonErrorResponseWithErrCode(w, http.StatusBadRequest, ErrCodeTunnelExist, "Tunnel already exist.")
 		return
 	}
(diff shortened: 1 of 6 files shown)


domenipavec (commented):
I don't think having client repository in the client is a good idea. I also don't think having methods that modify the clients are a good idea. Our main abstraction for interacting with clients is client service. I think the code for removing tunnels should live there.

morestatic:
i think i'm with domen here. 

we need to be clear and consistent regarding the responsibilities of our components. adding saving to the client (while fine if we'd started that way) is probably introducing inconsistencies in how we assign responsibilities. if we aren't consistent then the code will get harder to reason about and maintain.

my guess is that the tunnels code that's currently part of the client responsibilities should be moved up to the client services and longer term maybe even moved out to a tunnel services component that just focuses on managing tunnels.

i took a look at who's calling the client tunnel functions and it is really just the API layer and that has access to client services. i also looked at the implementation of the tunnel client functions and they aren't accessing lots of other client only functionality. so i don't think will be an issue to move that functionality up a layer, which will then make it easier and cleaner to save the client tunnels changes.

morestatic:
so ... i took a little risk and did the refactoring mentioned above.  i'm quite close to the client services and tunnels code at the moment, so it was pretty straightforward. personally, i feel this is better. what do you both think?

domenipavec:
@morestatic This is great.

Please uncomment and fix the tests.

morestatic:
Added the previously broken tests back in. They aren't failing now, so that's good.

domenipavec (approved):

thorstenkramm:
@morestatic Thanks for your quick solution. Much better than mine.
What about tunnels being closed by the tunnel scheduler or by inactivity? Is this saved to the db also?

PR #845, opened 2022-12-14
Fix/dev 2735 unrecoverable error

thorstenkramm:


share/logger/format.go
```diff
@@ -0,0 +1,13 @@
+package logger
+
+import (
+	"fmt"
+	"time"
+)
+
+func FormatConnectionState(disconnectedAt *time.Time) string {
```
domenipavec, line 8: This should probably be a method on client object.

client/client.go
```diff
@@ -261,15 +256,11 @@ func (c *Client) connectToMainOrFallback() (conn *sshClientConn, isPrimary bool,
...
-				continue
-			}
-			c.Errorf(fmt.Sprintf("Unrecoverable error: %s", err))
-			break
+			continue // Try the next server in the list
 		}
 		return conn, i == 0, nil
 	}
-	return nil, false, err
+	return nil, false, retryableError{err}
```
domenipavec, line 263: Is there any point of keeping a `retryableError` type or are we always going to retry on all errors? In that case we can probably just remove this type altogether.

domenipavec (approved):

PR #854, opened 2022-12-18
Extended logging

thorstenkramm:


client/client.go
```diff
@@ -190,10 +206,12 @@ func (c *Client) connectionLoop(ctx context.Context) {
 		if !isPrimary {
 			go func() {
 				for {
+					switchbackTimer := time.NewTimer(c.configHolder.Client.ServerSwitchbackInterval)
+					defer switchbackTimer.Stop()
```
morestatic, line 210: we should move this to an explicit stop after the `ctx.Done`

server/client_status_check.go
```diff
@@ -81,13 +81,15 @@ func (t *ClientsStatusCheckTask) PingClients(clientsToPing <-chan *clients.Clien
 		if !ok && err == nil && string(response) == "unknown request" {
 			t.log.Debugf("ping to %s [%s] succeeded in %s. client < 0.8.2", cl.Name, cl.ID, rtt)
 			cl.LastHeartbeatAt = &now
+			cl.DisconnectedAt = nil
```
morestatic, line 84: this should use `SetConnected`

server/client_status_check.go
```diff
@@ -81,13 +81,15 @@ func (t *ClientsStatusCheckTask) PingClients(clientsToPing <-chan *clients.Clien
...
 			results <- true
 			continue
 		}
 		// Only an empty response confirms the ping
 		if ok && err == nil && len(response) == 0 {
 			t.log.Debugf("ping to %s [%s] succeeded in %s. client >= 0.8.2", cl.Name, cl.ID, rtt)
 			cl.LastHeartbeatAt = &now
+			cl.DisconnectedAt = nil
```
morestatic, line 92: this should use `SetConnected`

server/clients/client_service.go
```diff
@@ -613,6 +613,7 @@ func (s *ClientServiceProvider) SetLastHeartbeat(clientID string, heartbeat time
 		return err
 	}
 	existing.LastHeartbeatAt = &heartbeat
+	existing.DisconnectedAt = nil
```
morestatic, line 616: this should use `SetConnected`

client/client.go
thorstenkramm: @morestatic Can you please do?

morestatic (approved):

PR #869, opened 2023-01-05
[DEV-2738] Server side configurable monitoring

morestatic:
https://tracker.rport.io/issue/DEV-2738/Switch-monitoring-off-less-than-1day-retention


rportd.example.conf
```diff
@@ -448,11 +448,15 @@
...
 
 [monitoring]
-  ## The rport server stores monitoring data of the clients for N days.
   ## https://oss.rport.io/advanced/monitoring/
-  ## Older data is purged automatically.
-  ## Default: 30 days
-  #data_storage_days = 30
+  ## Global switch to turn off monitoing system wide. Any monitoring settings on
+  ## client side will be ignored. Switched on by default.
+  # enabled = true
```
thorstenkramm, line 454: just for consistency. No blank space after the hash sign.
```
## Explanation double hash and blank
#sample_key = "sample_value"  
```

rportd.example.conf
```diff
@@ -448,11 +448,15 @@
...
+  ## The rport server stores monitoring data for a period of N.
+  ## Older data is purged automatically. Use suffix d (=days) or h (=hours)
+  ## Make sure to incldue quotes "" around the value
+  ## Default: "7d"
+  # data_storage_duration = "7d"
```
thorstenkramm, line 459: same

server/client_listener.go
```diff
@@ -397,17 +399,20 @@ func (cl *ClientListener) handleSSHRequests(clientLog *logger.Logger, clientID s
...
-				clientLog.Errorf("Failed to unmarshal save_measurement: %s", err)
-				continue
-			}
-			measurement.ClientID = clientID
-			err = cl.monitoringService.SaveMeasurement(context.Background(), measurement)
-			if err != nil {
-				clientLog.Errorf("Failed to save measurement for client %s: %s", clientID, err)
-				continue
+			// if server monitoring is disabled then do not save measurements even if received
+			if cl.Server.config.Monitoring.Enabled {
```
thorstenkramm, line 403: Shouldn't we better handle the disabled case and log an error and continue? (Monitoring enabled is still the default and not the exception.)
It's worth an error because some really strange must be going off if a client sends measurements with monitoring switched off.

rportd.example.conf
morestatic: done.

rportd.example.conf
morestatic: done.

server/client_listener.go
morestatic: ok. done.

server/server.go
```diff
@@ -300,9 +302,21 @@ func (s *Server) Run(ctx context.Context) error {
 	), s.config.Server.CheckClientsConnectionInterval)
 	s.Infof("Task to check the clients connection status will run with interval %v", s.config.Server.CheckClientsConnectionInterval)
 
-	cleaningPeriod := time.Hour * 24 * time.Duration(s.config.Monitoring.DataStorageDays)
-	go scheduler.Run(ctx, s.Logger, monitoring.NewCleanupTask(s.Logger, s.monitoringService, cleaningPeriod), cleanupMeasurementsInterval)
-	s.Infof("Task to cleanup measurements will run with interval %v", cleanupMeasurementsInterval)
+	if s.config.Monitoring.Enabled {
```
domenipavec, line 305: If we're keeping querying of data when monitoring is disabled, maybe we should also keep cleanup task so old data is cleaned up eventually?

server/chconfig/config.go
```diff
@@ -509,6 +524,59 @@ func matchingPorts(address1 string, address2 string) (matching bool, err error)
...
+		return err
+	}
+
+	if mc.Enabled && mc.GetDataStorageDuration() < time.Hour {
+		return errors.New("monitoring results must be stored for at least 1 hour")
+	}
+	return nil
+}
+
+func convertHourOrDayStringToDuration(desc string, inputStr string) (duration time.Duration, err error) {
```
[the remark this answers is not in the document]
morestatic: that's was my first thought and implementation ... but `time.Duration` doesn't like days.

server/server.go
morestatic: i checked with Thorsten. he said he'd prefer to disable the cleanup too. disabling monitoring is more of a troubleshooting assistance tool, so disabling is more like a freeze of the monitoring, so makes sense to disable the cleanup too. will add a note to the conf file though to indicate the cleanup will be paused when monitoring not running.

domenipavec (approved):

server/chconfig/config.go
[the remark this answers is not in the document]
domenipavec: Maybe add a comment explaining, so it's clear why the code is there.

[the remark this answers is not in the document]
morestatic: will do.

server/server.go
thorstenkramm: @morestatic But there is still a point we have overlooked. 
If monitoring is disabled, we should not accept API requests to the monitoring endpoints and on the `/status` we must add a field that informs about monitoring being enabled or disabled.

morestatic: I did wonder about that but thought that the user might be still interested in the measurements before the monitoring was paused? seems a little brutal to stop the viewing of historical data? agreed on adding a field to the `/status`.

thorstenkramm: @morestatic The most common use case for this feature are large installations where people can't provide big and fast enough disks. Monitoring will be disabled permanently. 
The data will be outdated very quickly.  Without disabling requests to the old monitoring data, we would have to rework the UI. People will always see latest data, but they will quickly overlook the date of the latest data.
Disabling the access to old data prevents confusion.

domenipavec (approved):

domenipavec (approved):

domenipavec (approved):

PR #873, opened 2023-01-09
Feature/dev 2683 mass deployment

edoardoc:
as of
https://tracker.rport.io/issue/DEV-2683/Mass-deployment-by-expirable-tokens


server/api_handler_plus_test.go
```diff
@@ -424,7 +423,7 @@ func setupTestAPIListenerForOAuth(
 	user := &users.User{
 		Username: "user1",
 		Password: "$2y$05$ep2DdPDeLDDhwRrED9q/vuVEzRpZtB5WHCFT7YbcmH9r9oNmlsZOm",
-		Token:    ptr.String("$2y$05$/D7g/d0sDkNSOh.e6Jzc9OWClcpZ1ieE8Dx.WUaWgayd3Ab0rRdxu"),
+		// Token:    ptr.String("$2y$05$/D7g/d0sDkNSOh.e6Jzc9OWClcpZ1ieE8Dx.WUaWgayd3Ab0rRdxu"),
```
domenipavec, line 426: Not sure why this is commented out. Either remove it or fix the test to accept it.

server/api_handler_me.go
```diff
@@ -89,6 +94,160 @@ func (al *APIListener) handleManageCurUserTotP(w http.ResponseWriter, req *http.
...
+			al.jsonErrorResponseWithTitle(w, http.StatusBadRequest, "missing body with scope.")
+			return
+		}
+
+		if !authorization.IsValidScope(r.Scope) {
+			al.jsonErrorResponseWithTitle(w, http.StatusBadRequest, "missing or invalid scope.")
+			return
+		}
+
+		if r.Scope == "clients-auth" && !user.IsAdmin() {
```
domenipavec, line 141: Why is `clients-auth` treated differently? Permissions will be checked when using the token anyway. And `read+write` token has a broader permission then `clients-auth`.

server/api_handler_me.go
```diff
@@ -89,6 +94,160 @@ func (al *APIListener) handleManageCurUserTotP(w http.ResponseWriter, req *http.
...
+		err = al.tokenManager.Save(req.Context(), updAPIToken)
+		if err != nil {
+			al.jsonError(w, err)
+			return
+		}
+
+		al.auditLog.Entry(auditlog.ApplicationAuthUserMeToken, auditlog.ActionUpdate).
+			WithHTTPRequest(req).
+			WithID(user.Username).
+			WithID(r.Prefix).
```
domenipavec, line 213: Can't have 2 id fields in auditlog.

server/api/authorization/model.go
```diff
@@ -0,0 +1,43 @@
...
+	"time"
+)
+
+const DefaultTimeoutSec = 60
+
+type APIToken struct {
+	Username  string     `json:"username,omitempty" db:"username"`
+	Prefix    string     `json:"prefix" db:"prefix"`
+	CreatedAt *time.Time `json:"created_at,omitempty" db:"created_at"`
+	ExpiresAt *time.Time `json:"expires_at,omitempty" db:"expires_at"`
```
domenipavec, line 15: ExpiresAt is not used during token validation?

server/api_handler_me.go
```diff
@@ -89,6 +94,160 @@ func (al *APIListener) handleManageCurUserTotP(w http.ResponseWriter, req *http.
...
+				Prefix: newPrefix,
+				Scope:  r.Scope,
+				Token:  newTokenClear,
+			}))
+	}
+
+	if action == "update" {
+		var r struct {
+			Prefix    string     `json:"prefix"`
+			ExpiresAt *time.Time `json:"expires_at"`
```
domenipavec, line 191: Why do we allow setting expires at with update but not during create?

server/api_handler_me.go
```diff
@@ -89,6 +94,160 @@ func (al *APIListener) handleManageCurUserTotP(w http.ResponseWriter, req *http.
...
+			WithHTTPRequest(req).
+			WithID(user.Username).
+			Save()
+
+		al.Debugf("APIToken %s is deleted for user [%s].", r.Prefix, user.Username)
+		w.WriteHeader(http.StatusNoContent)
+	}
+	if action == "create" {
+		var r struct {
+			Scope string `json:"scope"`
```
domenipavec, line 128: Why don't we allow changing scope with update? I think create and update should have the same fields.

server/api_handler_me.go
```diff
@@ -89,6 +94,160 @@ func (al *APIListener) handleManageCurUserTotP(w http.ResponseWriter, req *http.
...
+			authorization.APIToken{
+				Prefix: newPrefix,
+				Scope:  r.Scope,
+				Token:  newTokenClear,
+			}))
+	}
+
+	if action == "update" {
+		var r struct {
+			Prefix    string     `json:"prefix"`
```
domenipavec, line 190: Why allow changing prefix?
I don't think we should allow changing prefix. Especially not without some validation to disallow empty or same prefixes.

server/api/authorization/model.go
```diff
@@ -0,0 +1,43 @@
...
+	Scope     string     `json:"scope,omitempty" db:"scope"`
+	Token     string     `json:"token,omitempty" db:"token"`
+}
+
+func Extract(prefixedpwd string) (string, string, error) {
+	i := strings.Index(prefixedpwd, "_")
+	if i < 0 {
+		return "", "", errors.New("token should be in the format 'prefix_token'")
+	}
+	if i != 8 {
```
domenipavec, line 25: Prefix len should be a constant

server/api/authorization/model.go
```diff
@@ -0,0 +1,43 @@
+package authorization
+
+import (
+	"errors"
+	"strings"
+	"time"
+)
+
+const DefaultTimeoutSec = 60
```
domenipavec, line 9: What's this used for?

server/api/authorization/model.go
```diff
@@ -0,0 +1,43 @@
...
+
+func Extract(prefixedpwd string) (string, string, error) {
+	i := strings.Index(prefixedpwd, "_")
+	if i < 0 {
+		return "", "", errors.New("token should be in the format 'prefix_token'")
+	}
+	if i != 8 {
+		return "", "", errors.New("invalid token")
+	}
+	prefix := prefixedpwd[0:i]
```
domenipavec, line 28: I think it would be nicer to use `strings.SplitN` instead of index and then manually splitting.

server/api/authorization/sqlite.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+
+	"github.com/jmoiron/sqlx"
+
+	"github.com/cloudradar-monitoring/rport/share/query"
+)
+
+type SqliteProvider struct {
+	db        *sqlx.DB
+	converter *query.SQLConverter
```
domenipavec, line 15: Are you using converter anywhere?

server/api/authorization/sqlite.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, fmt.Errorf("unable to get api token from DB: %w", err)
+	}
+
+	return res, nil
+}
+
+func (p *SqliteProvider) Save(ctx context.Context, tokenLine *APIToken) (err error) {
+	res, err := p.db.NamedExecContext(
+		ctx,
+		"INSERT INTO"+
```
domenipavec, line 62: You could use ` for multiline query. You don't need to quote fields if they're simple.

server/api_handler_login_test.go
```diff
@@ -49,85 +222,103 @@ func TestPostToken(t *testing.T) {
...
-		Token:    ptr.String("$2y$05$/D7g/d0sDkNSOh.e6Jzc9OWClcpZ1ieE8Dx.WUaWgayd3Ab0rRdxu"),
 	}
 	userWithoutToken := &users.User{
 		Username: "user2",
 		Password: "$2y$05$ep2DdPDeLDDhwRrED9q/vuVEzRpZtB5WHCFT7YbcmH9r9oNmlsZOm",
-		Token:    nil,
 	}
+
+	// database for tokenManager, creates a token "read+write"
+	apiTokenDb, err := sqlite.New(":memory:", api_token.AssetNames(), api_token.Asset, DataSourceOptions)
```
domenipavec, line 273: I think most of this code is duplicated. You could have a helper function that initializes database and creates the token.

server/api_handler_commands_test.go
```diff
@@ -1122,247 +1119,6 @@ func TestHandlePostMultiClientCommandWithTags(t *testing.T) {
 	}
 }
 
-func TestHandlePostMultiClientWSCommandWithTags(t *testing.T) {
```
domenipavec, line 1125: Why delete unrelated test?

server/api_handler_commands_test.go
```diff
@@ -1590,250 +1346,6 @@ func TestHandlePostMultiClientScriptWithTags(t *testing.T) {
 	}
 }
 
-func TestHandlePostMultiClientWSScriptWithTags(t *testing.T) {
```
domenipavec, line 1593: Same

domenipavec (changes requested):

share/random/random.go
```diff
@@ -23,6 +23,11 @@ func AlphaNum(n int) string {
 	return String(n, alphaNum)
 }
 
+// ALPHANUM8 returns a random generated alphanum of 8 chars.
+var ALPHANUM8 = func() string {
```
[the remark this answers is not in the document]
edoardoc: I agree, its to support its redefinition during testing;
I didnt know how to redefine alphaNum

server/api_handler_plus_test.go
edoardoc: forgot; done

server/api_listener.go
```diff
@@ -393,11 +413,34 @@ func (al *APIListener) handleBasicAuth(username, password string) (authorized bo
...
+		return false, username, err
+	}
+
+	if userToken != nil {
+		tokenOk := verifyPassword(userToken.Token, password)
 		if tokenOk {
-			return true, username, nil
+			switch userToken.Scope {
+			case "read":
+				if httpverb == "GET" {
```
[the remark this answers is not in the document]
edoardoc: so this translates into:

if I want to handleBasicAuth with a valid token
AND
the token I provide is a "read" token
THEN
I am allowed only to access GET verb operations

server/api_handler_me.go
```diff
@@ -208,59 +367,31 @@ func (al *APIListener) handleGetIP(w http.ResponseWriter, req *http.Request) {
 	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload(ipResp))
 }
 
-type postTokenResponse struct {
-	Token string `json:"token"`
+func (al *APIListener) handleGetToken(w http.ResponseWriter, req *http.Request) {
+	al.handleManageCurUserAPIToken(w, req, "list")
```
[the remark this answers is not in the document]
edoardoc: could you point me to a place in the repo where I can find a pattern for that?

server/api_handler_login_test.go
edoardoc: indeed it is, I was going to ask you about how to handle this, ok will do;

server/api_handler_plus_test.go
edoardoc: done

server/api/authorization/model.go
edoardoc: leftover, removed

server/api_handler_me.go
edoardoc: prefix + username is the primary key, so it will never be updated: the only updateable field is ExpiresAt

server/api_handler_me.go
edoardoc: oups, 
I changed all of them in a more uniform format `WithID(fmt.Sprintf("[%s,%s]", user.Username, r.Prefix)).`

server/api_handler_me.go
edoardoc: that is now allowed, also added some testing.

server/api/users/user.go
```diff
@@ -7,7 +7,6 @@ type User struct {
 	PasswordExpired *bool    `json:"password_expired" db:"password_expired"`
 	Groups          []string `json:"groups" db:"-"`
 	TwoFASendTo     string   `json:"two_fa_send_to" db:"two_fa_send_to"`
-	Token           *string  `json:"token,omitempty" db:"token"`
```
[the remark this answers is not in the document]
edoardoc: yes, they will have to create new ones after upgrading

server/api_handler_me.go
```diff
@@ -89,6 +94,160 @@ func (al *APIListener) handleManageCurUserTotP(w http.ResponseWriter, req *http.
 	al.handleManageTotP(w, req, user, action)
 }
 
+func (al *APIListener) handleManageAPIToken(w http.ResponseWriter, req *http.Request, user *users.User, action string) {
+	if action == "delete" {
```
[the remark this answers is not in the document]
edoardoc: well maybe, but there is handleManageCurUserAPIToken that deals with the getUserModel;

server/api_handler_me.go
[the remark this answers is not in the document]
edoardoc: done, I made enums

server/api_handler_me.go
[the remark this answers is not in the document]
edoardoc: done

share/random/random.go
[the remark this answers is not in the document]
domenipavec: @edoardoc You will need to change `AlphaNum` to variable definition if you want to override it. Define it with `var AlphaNum = func` instead of `func AlphaNum`.

server/api_handler_login_test.go
```diff
@@ -183,6 +184,23 @@ func TestAPITokenOps(t *testing.T) {
...
+	ctx := context.Background()
+	itemToSave := authorization.APIToken{
+		Username:  username,
+		Prefix:    prefix,
+		CreatedAt: ptr.Time(time.Date(2001, 1, 1, 1, 0, 0, 0, time.UTC)),
+		ExpiresAt: ptr.Time(time.Date(2051, 1, 1, 2, 0, 0, 0, time.UTC)),
+		Scope:     scope,
+		Token:     token,
+	}
+	_ = dbProv.Save(ctx, &itemToSave)
```
domenipavec, line 218: You should check the error. You can pass testing.T as argument and use require.NoError here.

server/api_handler_me.go
```diff
@@ -117,7 +118,7 @@ func (al *APIListener) handleManageAPIToken(w http.ResponseWriter, req *http.Req
 		}
 		al.auditLog.Entry(auditlog.ApplicationAuthUserMeToken, auditlog.ActionDelete).
 			WithHTTPRequest(req).
-			WithID(user.Username).
+			WithID(fmt.Sprintf("[%s,%s]", user.Username, r.Prefix)).
```
domenipavec, line 123: I'm not sure this is great since this field could be used for filtering. I would just put `Username` here and pass prefix using `WithRequest(map[string]any{ "prefix": r.Prefix })`

server/api_handler_me.go
```diff
@@ -172,7 +173,7 @@ func (al *APIListener) handleManageAPIToken(w http.ResponseWriter, req *http.Req
 
 		al.auditLog.Entry(auditlog.ApplicationAuthUserMeToken, auditlog.ActionCreate).
 			WithHTTPRequest(req).
-			WithID(user.Username).
+			WithID(fmt.Sprintf("[%s,%s]", user.Username, newPrefix)).
```
domenipavec, line 180: Same.
Also pass all user input using `WithRequest`

server/api_handler_me.go
```diff
@@ -209,8 +209,7 @@ func (al *APIListener) handleManageAPIToken(w http.ResponseWriter, req *http.Req
 
 		al.auditLog.Entry(auditlog.ApplicationAuthUserMeToken, auditlog.ActionUpdate).
 			WithHTTPRequest(req).
-			WithID(user.Username).
-			WithID(r.Prefix).
+			WithID(fmt.Sprintf("[%s,%s]", user.Username, r.Prefix)).
```
domenipavec, line 217: Same

share/enums/scope.go
```diff
@@ -0,0 +1,9 @@
+package enums
```
domenipavec, line 1: `enums` package seems too generic for this. Why not just put it in `api` package?

server/api_handler_me.go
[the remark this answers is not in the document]
domenipavec: @edoardoc `handleManageCurUserAPIToken` has 9 lines, meanwhile this is a giant 150 line function. I would much prefer you put the code in each handler directly and duplicate those 9 lines.
Your way makes the code much harder to read. It also breaks all jump to definition navigation, since that throws you to the top of this giant function and then you need to figure out what action is.

server/api_handler_me.go
domenipavec: @edoardoc Your code is confusing, since you're reading prefix from the body it seems as if it's possible to update it.
Please make this the same as all other endpoints. Prefix is like id, so it should be part of the url for update (url should be '/me/token/{prefix}`).

server/api_handler_commands_test.go
domenipavec: DEV-2746 is for another repository (rportcli). Please address this in this PR, you can't just delete tests.

server/api_listener.go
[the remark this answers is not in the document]
domenipavec: @edoardoc Yes, but the problem is, that we have websocket (/ws) endpoints, that start as GET requests for authorization, but allow executing commands and scripts.

server/api_handler_commands_test.go
edoardoc: ok, sorry for this

server/api_handler_me.go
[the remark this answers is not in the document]
edoardoc: I was merely copying the same pattern used in other parts of this repo: anyway, done

server/api/authorization/model.go
edoardoc: done

share/enums/scope.go
edoardoc: moved it into authorization, afterall is its own package

server/api_handler_me.go
edoardoc: asked @thorstenkramm 

> read+write is not broader than client auth because read/write refers to the http verb and not to the resource.
read+write means you can read and write to any resource the user is allowed to access. Read+write as API token scope is not superseeding the resource permissions on user level.

server/api_listener.go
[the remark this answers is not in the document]
edoardoc: asked @thorstenkramm 

> There is a difference between API tokens and JWT. 
You can either go to /login fetch a JWT and then use this JWT for further requests.
Or you can query any API endpoint directly without going to the /login first by using an API Token.

server/api_handler_me.go
edoardoc: Verb is a PUT, so if I PUT a value (through the request body or not), which is part of the id, I inherently establish a change of the item that I am addressing: your view on the matter is syntactical only, but REST is semantics also; The approach with `/me/token/{prefix}` is certainly better 
...on it

server/api_handler_commands_test.go
edoardoc: done

server/api_handler_commands_test.go
edoardoc: tests are back and fixed with new APITokens

server/api/authorization/model.go
edoardoc: now it is

server/api_handler_me.go
```diff
@@ -89,6 +94,160 @@ func (al *APIListener) handleManageCurUserTotP(w http.ResponseWriter, req *http.
...
+		}
+		newPrefix := random.ALPHANUM8()
+
+		// token creation
+		tokenHash, err := bcrypt.GenerateFromPassword([]byte(newTokenClear), bcrypt.DefaultCost)
+		if err != nil {
+			al.jsonErrorResponse(w, http.StatusInternalServerError, err)
+			return
+		}
+		tokenHashStr := strings.Replace(string(tokenHash), users.HtpasswdBcryptAltPrefix, users.HtpasswdBcryptPrefix, 1)
```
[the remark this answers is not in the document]
edoardoc: refactored into users.GenerateTokenHash

server/api_handler_me.go
```diff
@@ -208,59 +214,193 @@ func (al *APIListener) handleGetIP(w http.ResponseWriter, req *http.Request) {
...
+			Scope:     r.Scope,
+			ExpiresAt: r.ExpiresAt,
+			Token:     newTokenClear,
+		}))
+}
+
+func (al *APIListener) handlePutToken(w http.ResponseWriter, req *http.Request) {
+	vars := mux.Vars(req)
+	prefix := vars[routes.ParamTokenPrefix]
+	if len(prefix) != 8 {
```
domenipavec, line 330: Use the constant

server/api_handler_login_test.go
domenipavec: @edoardoc You're saving 2 tokens, but only assigning and checking 1 error. Please check both errors.

server/api_listener.go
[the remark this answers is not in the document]
domenipavec: @edoardoc `read` token implies that user can only read data with this token. It should not allow executing scripts and commands through websocket endpoints.

domenipavec (commented):
I unresolved a few comments above. Please also address those.

server/api_listener.go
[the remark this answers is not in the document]
edoardoc: done

server/api/authorization/sqlite.go
edoardoc: nope, leftover, removed

server/api_handler_login_test.go
edoardoc: CommonAPITokenTestDb creates always two tokens in the db, 

first one with the provided params,
second one with the provided params replacing prefix with "expired1" and ExpiresAt = 2001;
most of the tests they check only for the first one, 

https://github.com/cloudradar-monitoring/rport/blob/69e562c7a51325ab4fce619c63a21958ec6b1469/server/api_handler_login_test.go#L376

checks for the second one.

server/api_handler_me.go
edoardoc: done

edoardoc:
> I unresolved a few comments above. Please also address those.

I did what I could trying to understand which ones are still standing.
please add a simple "this" comment to the ones you believe they still need addressing

server/api/authorization/model.go
domenipavec: @edoardoc This (but I guess if you feel strongly your way is better I'm also ok keeping it as is).

server/api_handler_login_test.go
```diff
@@ -19,6 +25,203 @@ import (
...
+		Username:  username,
+		Prefix:    "expired1",
+		CreatedAt: ptr.Time(time.Date(2001, 1, 1, 1, 0, 0, 0, time.UTC)),
+		ExpiresAt: ptr.Time(time.Date(2001, 1, 1, 2, 0, 0, 0, time.UTC)),
+		Scope:     scope,
+		Token:     token,
+	}
+	_ = dbProv.Save(ctx, &itemToSave)
+
+	require.NoError(t, err)
```
domenipavec, line 221:
```suggestion
	itemToSave := authorization.APIToken{
		Username:  username,
		Prefix:    prefix,
		CreatedAt: ptr.Time(time.Date(2001, 1, 1, 1, 0, 0, 0, time.UTC)),
		ExpiresAt: ptr.Time(time.Date(2051, 1, 1, 2, 0, 0, 0, time.UTC)),
		Scope:     scope,
		Token:     token,
	}
	err = dbProv.Save(ctx, &itemToSave)
	require.NoError(t, err)

	itemToSave = authorization.APIToken{
		Username:  username,
		Prefix:    "expired1",
		CreatedAt: ptr.Time(time.Date(2001, 1, 1, 1, 0, 0, 0, time.UTC)),
		ExpiresAt: ptr.Time(time.Date(2001, 1, 1, 2, 0, 0, 0, time.UTC)),
		Scope:     scope,
		Token:     token,
	}
	err = dbProv.Save(ctx, &itemToSave)
	require.NoError(t, err)

```

server/api/authorization/model.go
edoardoc: honestly cant see which one is better :)
but here you are...

server/api/authorization/model.go
```diff
@@ -0,0 +1,52 @@
...
+	i := strings.Index(prefixedpwd, "_")
+	if i < 0 {
+		return "", "", errors.New("token should be in the format 'prefix_token'")
+	}
+	if i != APITokenPrefixLength {
+		return "", "", errors.New("invalid token")
+	}
+	parts := strings.SplitN(prefixedpwd, "_", i)
+	prefix := parts[0]
+	token := parts[1]
```
domenipavec, line 39:
```suggestion
	parts := strings.SplitN(prefixedpwd, "_", 2)
	if len(parts) < 2 {
		return "", "", errors.New("token should be in the format 'prefix_token'")
	}
	prefix := parts[0]
	token := parts[1]
	if len(prefix) != APITokenPrefixLength {
		return "", "", errors.New("invalid token")
	}
```

domenipavec (approved):

PR #893, opened 2023-01-19
Issue/dev 2732/make tls version configurable

edoardoc:
https://tracker.rport.io/issue/DEV-2732/make-TLS-version-configurable


cmd/rportd/main.go
```diff
@@ -346,6 +346,7 @@ func init() {
 	viperCfg.SetDefault("api.totp_account_name", "RPort")
 	viperCfg.SetDefault("api.password_min_length", 14)
 	viperCfg.SetDefault("api.password_zxcvbn_minscore", 0)
+	viperCfg.SetDefault("api.tls_min", "1.3")
```
domenipavec, line 349: Why set default for one but not the other? Please do the same for both.
And since you're treating empty values as 1.3 in the code, you don't need the default here (or you can set the default for both here and remove empty string cases in the code).

server/api_listener.go
```diff
@@ -195,7 +196,14 @@ func NewAPIListener(
 
 	userService := users.NewAPIService(usersProvider, config.API.IsTwoFAOn(), config.API.PasswordMinLength, config.API.PasswordZxcvbnMinscore)
 
-	HTTPServerOptions := []chshare.ServerOption{chshare.WithTLS(config.API.CertFile, config.API.KeyFile, security.TLSConfig)}
+	tlsMin := uint16(tls.VersionTLS13)
+	if config.API.TLSMin != "" && config.API.TLSMin != "1.3" {
```
domenipavec, line 200: Use `switch` statement to make this more readable.

server/clients/clienttunnel/tunnel_internal_proxy.go
```diff
@@ -122,7 +126,16 @@ func (tp *InternalTunnelProxy) Start(ctx context.Context) error {
 
 func (tp *InternalTunnelProxy) listen() {
 	tp.Logger.Debugf("listener starting")
-	tp.proxyServer.TLSConfig = security.TLSConfig
+
+	// this tlsmin is the InternalTunnelProxyConfig config in the server section
+	tlsMin := uint16(tls.VersionTLS13)
+	if tp.Config.TLSMin != "" && tp.Config.TLSMin != "1.3" {
```
domenipavec, line 132: Same as above, use switch statement.
Also maybe extract common code here and above in a helper function?

cmd/rportd/main.go
```diff
@@ -371,6 +372,7 @@ func bindPFlags() {
 	_ = viperCfg.BindPFlag("server.allow_root", pFlags.Lookup("allow-root"))
 	_ = viperCfg.BindPFlag("server.tunnel_proxy_cert_file", pFlags.Lookup("tunnel-proxy-cert-file"))
 	_ = viperCfg.BindPFlag("server.tunnel_proxy_key_file", pFlags.Lookup("tunnel-proxy-key-file"))
+	_ = viperCfg.BindPFlag("server.tls_min", pFlags.Lookup("tls_min"))
```
domenipavec, line 375: Since this would only set one `tls_min` value, I would avoid making a flag at all. I think it's fine and less confusing if both values are only settable through config file.
Alternatively you could create `api-tls-min` and `proxy-tls-min` flags.

rportd.example.conf
```diff
@@ -211,6 +211,11 @@
   ## Maximum number of results to keep for commands, scripts and schedules execution
   #jobs_max_results = 10000
 
+  ## Minimal TLS version required for Internal Tunnel
+  ## Default 1.3
+  ## Possible settings: 1.3 or 1.2
+  # tls_min = "1.3"
```
domenipavec, line 217: Remove space after # in example setting.

rportd.example.conf
```diff
@@ -355,6 +360,11 @@
   ## 4 strong protection from offline attacks under same scenario: "very unguessable"
   ## -1 zxcvbn check is disabled
   #password_zxcvbn_minscore = 0
+  
+  ## Minimal TLS version required for the API
+  ## Default 1.3
+  ## Possible settings: 1.3 or 1.2
+  # tls_min = "1.3"
```
domenipavec, line 367: Same.

cmd/rportd/main.go
edoardoc: set both here

server/clients/clienttunnel/tunnel_internal_proxy.go
edoardoc: yes, using the string and the switch in same function

cmd/rportd/main.go
edoardoc: removed the parameter

share/security/tls_config.go
```diff
@@ -2,8 +2,18 @@ package security
...
 import "crypto/tls"
 
-var TLSConfig = &tls.Config{
-	MinVersion:               tls.VersionTLS13,
-	CurvePreferences:         []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
-	PreferServerCipherSuites: true,
+func TLSConfig(configTLSMin string) *tls.Config {
+	tlsMin := uint16(tls.VersionTLS13)
+	switch configTLSMin {
+	case "1.2":
```
domenipavec, line 8: Nit.: Switch seems a bit excessive now that's only one case, but feel free to leave it as is.

domenipavec (approved):

PR #895, opened 2023-01-25
Issue/dev 2641/groups based on two or more tag filters

edoardoc:
https://tracker.rport.io/issue/DEV-2641


share/query/matches_filter.go
```diff
@@ -40,32 +41,38 @@ func matchesFilter(valueMap map[string]interface{}, filter FilterOption) (bool,
...
+					nMatches++
 				}
 			}
 		}
 	}
 
-	return false, nil
+	switch filter.ValuesLogicalOperator {
+	case FilterLogicalOperatorTypeAND:
+		return nMatches == len(filter.Values), nil
```
domenipavec, line 73: I don't think this code is correct, because you could have multiple values (for multiple columns, or columns with multiple values) that match the same filter value and still get `nMatches == len(filter.Values)`. (Example filter values `linux` and `a*`, would match tags `abc` and `ade`)

The solution I think is to move the `for _, filterValue := range filter.Values`  and make it the outer most loop. Then when you get the match inside, you need to do a `continue` for the outer most loop.
That way each `filterValue` will be able to match only once, and then this comparison will work correctly.

domenipavec (changes requested):

share/query/matches_filter.go
edoardoc: these changes are only to support the condition **FilterLogicalOperatorTypeAND** and for the fields that support multiple values/ filters (tags).
You are right about the multiple matches - I am isolating some test cases to sort out the problem. 
Your solution is not applicable though, cause cycling the filters in the external cycle would establish a priority among them: if filter = { "*a*", "Datacenter 2" } and tags is { "Datacenter 2", "Linux" } then the second filter will never be tested cause the match will occur always with the first one.
Probably we will need a data structure for the filters to be emptied during the loops.

edoardoc: ended up using a map to avoid reusing filters in the same cycle.
let me know if you find a case where this would fail...

share/query/matches_filter.go
```diff
@@ -26,6 +26,8 @@ func MatchesFilters(v interface{}, filterOptions []FilterOption) (bool, error) {
 }
 
 func matchesFilter(valueMap map[string]interface{}, filter FilterOption) (bool, error) {
+	matches := map[string]string{}
```
domenipavec, line 29: Nit.: This could be a map to bools, no need to store strings. Also you could preallocate the right size map (`make(map[string]bool, len(filter.Values))`).

domenipavec (approved):
I made a small optimization improvement suggestion, otherwise this is great.

share/query/matches_filter.go
edoardoc: yes, two good points, done!

domenipavec (approved):

PR #896, opened 2023-01-26
Fix/dev 2757 broken plugin

thorstenkramm:
@morestatic I've created that branch to fix the plus plugin building. It now contains a workflow that tests the plugin building and loading against each push to make sure we always have a running combination of `rportd` and `rport-plus.so`.

Please fix what's needed to make the workflow succeed.


morestatic:
hi thorsten, i think the workflow is now passing. could you double-check meeting expectations from your perspective and then if you're happy i'll merge the PR.

PR #921, opened 2023-02-23
Issue/dev 2773/api token management improvement

edoardoc:
https://tracker.rport.io/issue/DEV-2773/api-token-management-improvement


db/migration/api_token/sql/001_init.up.sql
```diff
@@ -1,9 +1,13 @@
 CREATE TABLE api_token (
```
thorstenkramm, line 1: If the endpoint is plural, the table must use plural too. `api_tokens`.

db/migration/api_token/sql/001_init.up.sql
```diff
@@ -1,9 +1,13 @@
...
     created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP, 
     expires_at DATETIME,
     scope TEXT,
     token TEXT NOT NULL,
     PRIMARY KEY (username, prefix)
-) WITHOUT ROWID;
\ No newline at end of file
+) WITHOUT ROWID;
+
+CREATE UNIQUE INDEX api_token_unique_name
```
thorstenkramm, line 12: see above

api-doc/openapi/paths/me_token.yaml
```diff
@@ -53,6 +53,9 @@ put:
             prefix:
```
thorstenkramm, line 53: The API shall return just the token, including the prefix. The prefix is useless and irrelevant for the user.

thorstenkramm (changes requested):

db/migration/api_token/sql/001_init.up.sql
```diff
@@ -1,9 +1,13 @@
 CREATE TABLE api_token (
     username TEXT NOT NULL CHECK (username != ''),
     prefix TEXT NOT NULL CHECK (prefix != ''),
+    name TEXT NOT NULL,
```
domenipavec, line 4: This migration won't be run for existing databases. You should create a new migration that adds a column.

server/api_handler_login_test.go
```diff
@@ -174,7 +210,7 @@ func TestAPITokenOps(t *testing.T) {
 
 			// then
 			require.Equal(tc.wantStatusCode, w.Code)
-			if tc.wantErrTitle == "" {
+			if strings.HasPrefix(fmt.Sprintf("%d", tc.wantStatusCode), "2") { // any 2** status code is a success test case
```
domenipavec, line 213: This is a bit weird way to test this, you could just do `tc.wantStatusCode/100 == 2`.

server/api_handler_login_test.go
```diff
@@ -183,16 +219,25 @@ func TestAPITokenOps(t *testing.T) {
...
+
+				// decode response
+				tResponse := api.NewErrAPIPayloadFromMessage("", "", "")
+				err = json.Unmarshal(w.Body.Bytes(), &tResponse)
 				require.NoError(err)
-				require.Equal(string(wantRespBytes), w.Body.String())
+
+				// compare items of struct error, one by one,
+				if tc.wantErrTitle != "" {
+					require.Equal(tc.wantErrTitle, tResponse.Errors[0].Title)
```
domenipavec, line 230: You should use `assert.Equal` so the test continues if this is not equal and we can see all failures.
`require` should be used in cases when it makes no sense to continue, like above if json unmarshal fails with errors it makes no sense to do these comparisons.

domenipavec (changes requested):

db/migration/api_token/sql/001_init.up.sql
edoardoc: Entities should be named in a singular way - especially database ones and regardless whatever bs one does on the next layer.
I could speak about this for a while, though checking the codebase I understand there is the same mistake in all other tables so I will just follow the convention here.

server/api_handler_login_test.go
edoardoc: oups :)

server/api_handler_me.go
```diff
@@ -277,11 +285,21 @@ func (al *APIListener) handlePostToken(w http.ResponseWriter, req *http.Request)
...
+	}
+
 	if r.Scope == authorization.APITokenClientsAuth && !user.IsAdmin() {
 		al.jsonErrorResponseWithTitle(w, http.StatusBadRequest, "current user should belong to Administrators group to create a token with this scope")
 		return
 	}
 
+	createdAt := ptr.Time(time.Now().Truncate(time.Second).UTC())
+	if r.ExpiresAt == nil {
+		r.ExpiresAt = ptr.Time(createdAt.AddDate(1, 0, 0))
```
domenipavec, line 300: Is this a requirement to not have tokens without expiration date?

Can you add a comment so it's clear what is getting added? Like `AddDate(1 /* year */, 0, 0)`

edoardoc: yes it is a requirement, default for expiry date is 1 year from creation date (if not provided)
I will add a comment

PR #984, opened 2023-03-28
Dev 2641

edoardoc:


server/cgroups/group.go
```diff
@@ -65,17 +65,42 @@ type ClientParams struct {
...
+			for k, curParamsOperands := range curGenericParam {
+				matches := 0
+				for _, curValue := range values {
+					for _, curParamAnd := range curParamsOperands {
+						if Param(curParamAnd).matches(curValue) {
+							matches++
+						}
+					}
+					switch strings.ToLower(k) { // operators
+					case "and":
```
domenipavec, line 87: What happens if the map has both `and` and `or`? Is that a documented behavior?

server/clients/client_test.go
```diff
@@ -128,6 +128,40 @@ func TestClientBelongsToGroup(t *testing.T) {
...
 		},
+		{
+			name: "match with new and",
+
+			client: c1,
+			group: &cgroups.ClientGroup{
+				ID: "group-1",
+				Params: &cgroups.ClientParams{
+					// OR
+					// Tag: &cgroups.ParamValues{"Li nux", "Datacenter 2"},
```
domenipavec, line 139: I'm not sure about these comments. Is this something to be deleted or is it supposed to mean something?

server/cgroups/group.go
```diff
@@ -65,17 +65,42 @@ type ClientParams struct {
...
 	}
-
-	for _, curParam := range *p {
-		for _, curValue := range values {
-			if curParam.matches(curValue) {
-				return true
+	var curParam Param
+	for _, curGenericParam := range *p {
+		switch curGenericParam := curGenericParam.(type) {
+		case map[string][]string:
```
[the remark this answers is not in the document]
edoardoc: I did some changes, now it should work

server/cgroups/group.go
```diff
@@ -65,17 +65,42 @@ type ClientParams struct {
...
+			for k, curParamsOperands := range curGenericParam {
+				matches := 0
+				for _, curValue := range values {
+					for _, curParamAnd := range curParamsOperands {
+						if Param(curParamAnd).matches(curValue) {
+							matches++
```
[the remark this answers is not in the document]
edoardoc: changed the matches strategy into a map, also added some tests for this case

server/cgroups/group.go
edoardoc: no, I will contact @thorstenkramm for this

server/clients/client_test.go
edoardoc: removed

server/cgroups/group.go
thorstenkramm: @edoardoc @domenipavec 
Are you referring to an input like this?
```
"params": {
   "tags": {"and" : ["A", "B"], "or": ["C","D"] }
}
```
This must be denied. If the length of the array exceeds 1, we must exit with an error.

edoardoc: @thorstenkramm 
the right input would be:

```
"params": {
    "tag": [ { "and": ["T*", "Datacenter 2", "Datacenter 5"] } ]
}

```
tag is an array of strings (old specs) or an array of maps (new specs)
I am adding a test case in the [testfile](https://github.com/realvnc-labs/rport/blob/046efe305e08397039a5cb8999070c5a3ad09518/server/clients/client_test.go#L302) to address the situation

server/clients/client_test.go
```diff
@@ -267,6 +268,99 @@ func TestClientBelongsToGroup(t *testing.T) {
...
+		Version:      "0.1.12",
+		Address:      "88.198.189.163:50078",
+		ClientAuthID: "client-auth-1",
+	}
+
+	var andTestCaseFalse cgroups.ClientParams
+	err := json.Unmarshal([]byte(`{ "tag": [{ "and": ["Linux", "Datacenter 2"] }] }`), &andTestCaseFalse)
+	assert.NoError(t, err)
+
+	var andTestCaseTrue cgroups.ClientParams
```
domenipavec, line 292: This is a bit weird way to write test cases, I would just put json in the  test cases struct and unmarshal in the loop.

server/cgroups/group.go
```diff
@@ -105,6 +105,55 @@ func (p Param) matches(value string) bool {
...
+		operands = curGenericParam[operator]
+	} else {
+		// unmarshaling as "and|or" : [ "first", "second"] failed
+		var listPattern []string
+		err := json.Unmarshal(*p, &listPattern)
+		if err == nil {
+			matches = make(map[string]bool, len(listPattern))
+			operands = listPattern
+		} else {
+			// unmarshaling as [ "first", "second"] failed
```
domenipavec, line 129: I think you should split the parsing in a separate fanction, that can then be used for validation when `Params` are submitted through the api and this error reported to the user.

server/clients/client_test.go
edoardoc: changed

server/cgroups/group.go
edoardoc: now it works with

```
"params": {
    "tags": { "and": ["T*", "Datacenter 2", "Datacenter 5"] }
}
```

and 
```

"params": {
    "tags": [ "T*", "Datacenter 2", "Datacenter 5" ]
}
```

server/api_handler_clientgroups.go
```diff
@@ -100,6 +102,25 @@ func validateInputClientGroup(group cgroups.ClientGroup) error {
 	if invalidGroupIDRegexp.MatchString(group.ID) {
 		return fmt.Errorf("invalid group ID %q: can contain only %q", group.ID, validGroupIDChars)
 	}
+	if group.Params != nil {
+		var curGenericParam map[string][]string
```
domenipavec, line 106: This parsing is mostly the same as in the `MatchesRawTags`, you should extract the parsing into one function and use it in both places, otherwise it's dangerous that it will get out of sync with further changes.

server/api_handler_clientgroups.go
```diff
@@ -100,6 +102,25 @@ func validateInputClientGroup(group cgroups.ClientGroup) error {
...
+			// unmarshaling as "and|or" : [ "first", "second"] failed
+			var listPattern []string
+			err := json.Unmarshal(*group.Params.Tag, &listPattern)
+			if err != nil {
+				// unmarshaling as [ "first", "second"] failed
+				return errors.New("error parsing tags group definitions")
+			}
+		} else {
+			// operator check
+			operator := strings.ToLower(reflect.ValueOf(curGenericParam).MapKeys()[0].String())
```
domenipavec, line 118: What happens if there are no map keys, probably panic? You have a check for this in the code below but not here. Exactly the reason I think this should just be merged into one function.
Also I think Thorsten wanted an error if both `and` and `or` are present.

server/cgroups/group.go
```diff
@@ -105,6 +105,58 @@ func (p Param) matches(value string) bool {
...
+		err := json.Unmarshal(*p, &listPattern)
+		if err == nil {
+			matches = make(map[string]bool, len(listPattern))
+			operands = listPattern
+		} else {
+			return false // also unmarshaling as [ "first", "second"] failed
+		}
+	}
+	if len(operands) == 0 {
+		return false
```
domenipavec, line 133: This is a bit weird, I would expect an always match with no values (same as on line 110).

server/clients/client_test.go
```diff
@@ -155,41 +154,9 @@ func TestClientBelongsToGroup(t *testing.T) {
 				Description: "Group with no params",
 				Params:      &cgroups.ClientParams{},
 			},
-
 			wantRes: false,
 		},
-		{
```
domenipavec, line 161: Why are these test cases removed, shouldn't you just replace tag with `dataEmpty`?

domenipavec (changes requested):

server/clients/client_test.go
edoardoc: image

I think these are both wrong:

- the first one, "group with no tags, client with nil tags" gives a match for Tag = empty when data.tags == nil 

- the second one, "group with no tags, client with no tags" gives a match for Tag = empty when data.tags is empty

I believe the best outcome would be to give an error when Tag = empty or nil and then compare only non empty tags.

server/cgroups/group.go
```diff
@@ -105,6 +105,76 @@ func (p Param) matches(value string) bool {
...
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func ParseTag(p *json.RawMessage) (string, []string, error) {
+	operator := "or" // default
```
domenipavec, line 145: Nit.: I think operator strings should be defined as constants.

server/clients/client_test.go
domenipavec: There are a few cases of client param tags that I think we should cover, what client tags are shouldn't matter for these cases:

- When `tag` key is not present in `params`. This should be `true` and I think in this case `*json.RawMessage` is nil and it's covered by one of the test cases.
- When `"tag"=null`. I think this is the same as above, but maybe in this case `json.RawMessage = "null"`, and it won't parse correctly?
- When `"tag"=[]` or `"tag"={"or":[]}`. The behavior of `MatchesOneOf` for this is to always return `true`. I think we should match this behavior to be consistent.
- When `"tag"={"and":[]}`. I think this should be the same as for `or`, but it could be argued that it could also always be `false`. I think I would leave it up to Thorsten.

edoardoc: for `tag = null` there is "group with unset tags, client with tags"
for `tag = []` or `tag = { anything }` indeed `MatchesOneOf` returns true, - erroneously: why there should be a match if there is no tags coming in? the same goes for `and`.
All these conditions should be caught and tested by `validateInputClientGroup` (checking the error condition of `ParseTag`) but you are right, it seems there is no validation for whatever goes in `MatchesOneOf`, neither are error messages produced...

domenipavec: I'll approve for now, the behavior of these corner cases is up to Thorsten, we can also change it later if required.

domenipavec (approved):

PR #989, opened 2023-03-30
[DEV-2769] Acme for api and tunnel proxy

domenipavec:
https://tracker.rport.io/issue/DEV-2769/built-in-lets-encrypt-acme


server/acme/acme.go
```diff
@@ -0,0 +1,83 @@
...
+		Logger: l,
+		hosts:  make(map[string]bool),
+	}
+	a.manager = &autocert.Manager{
+		Cache:      autocert.DirCache(filepath.Join(dataDir, "acme")),
+		Prompt:     autocert.AcceptTOS,
+		HostPolicy: a.hostPolicy,
+	}
+	if httpPort > 0 {
+		go a.listenHTTP(httpPort)
```
morestatic, line 35: i'm not a big fan of starting go-routines (and similar activities) in constructors. it means that the initialisation and the starting are tightly coupled, which quite often isn't what we want. can we add a `Start` fn instead?

server/acme/acme.go
```diff
@@ -0,0 +1,83 @@
...
+	return a
+}
+
+func (a *Acme) listenHTTP(httpPort int) {
+	addr := fmt.Sprintf(":%d", httpPort)
+	a.Infof("listening on %s", addr)
+
+	server := &http.Server{
+		Addr:              addr,
+		ReadHeaderTimeout: 3 * time.Second,
```
morestatic, line 46: are there any scenarios where we'd like these to be different from 3 secs? with the concurrency work, there was quite a bit of work to find all these constant timeouts. even though only one use, i'd prefer this to be a constant at the top of the file (or somewhere similar), so that we don't have to search through the code to realise it is here.

share/http_server.go
```diff
@@ -41,7 +41,10 @@ func NewHTTPServer(maxHeaderBytes int, l *logger.Logger, options ...ServerOption
 		httpLogger = l.Fork("http-server")
 	}
 	s := &HTTPServer{
-		Server:   &http.Server{MaxHeaderBytes: maxHeaderBytes, ReadHeaderTimeout: 5 * time.Second},
+		Server: &http.Server{
+			MaxHeaderBytes:    maxHeaderBytes,
+			ReadHeaderTimeout: 5 * time.Second,
```
morestatic, line 46: again, can we make this a constant and move the top of the file (or at least somewhere easy to see). maybe next to `maxHeaderBytes` declaration?

server/acme/acme.go
domenipavec: Done

server/acme/acme.go
domenipavec: This server is only for acme validation and this is the recommended value. Moved to a constant.

share/http_server.go
domenipavec: Moved to a constant.
`maxHeaderBytes` comes from the config file, so not next to that.

morestatic (approved):

PR #1001, opened 2023-04-06
[DEV-2836] Partially unresponsive clients after server restart

morestatic:
This fixes an issue where the connection was breaking, so there was another round of the connection loop. However the connection was still marked as "active", so the SSH request handling go-routine wasn't restarted. The fix just marks the connection as no longer active on a connection break.
Also, the PR includes some additional debug logs. Particularly, the request types and payloads received from the server are now logged. This should give some clues as to what request a client might be processing at the time of client issues.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -304,6 +304,7 @@ func (c *Client) connectionLoop(ctx context.Context, withInitialSendRequestDelay
 
 		c.Logger.Infof("connection wait stopped")
 
+		c.setConnActive(false)
 		c.setConn(nil)
 		c.updates.SetConn(nil)
 		c.monitor.SetConn(nil)
@@ -329,16 +330,19 @@ func (c *Client) handleConnectionError(backoff *backoff.Backoff, connerr error) 
 	c.showConnectionError(connerr, attempt)
 
 	if errors.Is(connerr, context.Canceled) {
+		c.Errorf("context canceled")
 		return true
 	}
 
 	// check if the user has set a max retry limit
 	if c.configHolder.Connection.MaxRetryCount >= 0 && attempt >= c.configHolder.Connection.MaxRetryCount {
+		c.Errorf("max retries exceeded")
 		return true // if so, stop trying
 	}
 
 	var d = backoff.Duration()
 	if _, ok := connerr.(comm.TimeoutError); ok {
+		c.Debugf("reseting backoff timer")
 		// Timeout means the server isn't offline, so reset the backoff and use an initial short retry duration
 		backoff.Reset()
 		rand.Seed(time.Now().UnixNano())
@@ -516,13 +520,14 @@ func (c *Client) sendConnectionRequest(ctx context.Context, sshConn ssh.Conn, mi
 
 	if err != nil {
 		c.Errorf("connection request err = %v", err)
+		c.Errorf("closing sshConn")
 		if closeErr := sshConn.Close(); closeErr != nil {
 			c.Errorf("Failed to close connection: %s", closeErr)
 		}
 		reconnect := strings.Contains(err.Error(), "reconnect")
 		if reconnect {
 			reconnectDelay := ServerReconnectRequestBackoffTime + (time.Duration(rand.Intn(30)) * time.Second)
-			c.Debugf("waiting %d seconds before reconnect", reconnectDelay/time.Second)
+			c.Debugf("reconnect requested. waiting %d seconds before retrying.", reconnectDelay/time.Second)
 			// this probably means the server is too busy for us. wait quite a while
 			// before returning to the conn loop.
 			time.Sleep(reconnectDelay)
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -304,6 +304,7 @@ func (c *Client) connectionLoop(ctx context.Context, withInitialSendRequestDelay
 
 		c.Logger.Infof("connection wait stopped")
 
+		c.setConnActive(false)
```
domenipavec, line 307: @morestatic Why do we need to track conn active state? How is it different then just checking `c.sshConnection != nil`?
It seems this is a state of the connection being active not of the client having active connection. Is that correct?
Maybe we should then wrap connection in a struct that has this property, so this state could be tracked per connection?

morestatic: bottom line, i think we need to rewrite the client connection handling ...

however, until then, we have a scenario when the server is busy so doesn't respond to the connection request from the client, but there's nothing wrong with the ssh connection itself. this is a non-active connection. once the server responds then we start the ssh "rport server request" handler go-routine (with the already working ssh connection) and mark the connection as active.

ideally, we don't want to close the ssh connection when the server times out, because this could happen with many clients at similar times, potentially causing a mini-herd to rush the server again with new ssh connections (when there wasn't anything wrong with the old ones). btw, this maybe less of an issue now with the two step connection handling on the server but that would need investigation.

i really don't like the way the client hasn't separated the concerns of the connection state from exchanging messages with the server and given 1.0 is supposed to be v.soon now, it doesn't seem appropriate to rewrite at this time. therefore it feels like we're a bit stuck with the current (potentially fragile) solution.

however, we need the current solution to work, which is one of the reason that I created the PR and requested your review for a one line fix. can you use your eagle eyes to see if you can see any other bugs or issues that could hurt connections / re-connections?

domenipavec: I don't think that what you're describing is correct, because `setConnActive(true)` happens on line 265 before the connection request is sent on line 278.

Also the only place where this value is used in on line 260. But on that line we always have a fresh connection from the select statement, either through switchback connect or through `connectToMainOrFallback`. So on line 260 the value should always be `false`. So having this value makes no sense and we should be able to remove it unless I'm missing something?

morestatic: thanks for the eagle eyes. indeed the current code doesn't look correct.

it is late here now. will take a another look in the morning with a fresh mind and revert. if we can simply remove it then that'll be good. i'm a little worried it isn't that simple, but need a clear head to be sure.

morestatic: ok. so, i've removed the active connection concept. been running load and restart tests for most of the day, and no obvious issues. mass clients do seem a bit slower to reconnect but that won't be related. we have change the server specs since some of the original testing. things seem to be reliable at the moment, so i won't make anymore tweaks.

can you take another quick look when you get a chance and make sure all seem fairly sensible?

client/client.go
```diff
@@ -329,16 +326,19 @@ func (c *Client) handleConnectionError(backoff *backoff.Backoff, connerr error)
 	c.showConnectionError(connerr, attempt)
 
 	if errors.Is(connerr, context.Canceled) {
+		c.Errorf("context canceled")
 		return true
 	}
 
 	// check if the user has set a max retry limit
 	if c.configHolder.Connection.MaxRetryCount >= 0 && attempt >= c.configHolder.Connection.MaxRetryCount {
+		c.Errorf("max retries exceeded")
```
domenipavec, line 335: Same.

client/client.go
```diff
@@ -171,20 +171,24 @@ func (c *Client) setConn(sshConnection ssh.Conn) {
...
-	defer c.mu.RUnlock()
-	return c.connActive
-}
-
-func (c *Client) setConnActive(active bool) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	c.connActive = active
+func printMemStats(c *Client, rtm runtime.MemStats) {
+	runtime.GC()
```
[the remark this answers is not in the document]
morestatic: ok. yes. i did mean to remove the GC. removed and tidied up the code a little.

client/client.go
```diff
@@ -171,20 +171,24 @@ func (c *Client) setConn(sshConnection ssh.Conn) {
...
+		rtm.NumGC,
+		time.UnixMilli(int64(rtm.LastGC/1_000_000)),
+	)
 }
 
 func (c *Client) keepAliveLoop(ctx context.Context) {
+	var rtm runtime.MemStats
+
 	for c.isRunning() {
+		runtime.ReadMemStats(&rtm)
```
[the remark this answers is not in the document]
morestatic: i guess it depends how often the memory needs to be allocated and de-allocated, but in this scenario it is low use so better to contain the stats in the print fn. updated.

client/client.go
```diff
@@ -329,16 +326,19 @@ func (c *Client) handleConnectionError(backoff *backoff.Backoff, connerr error)
 	c.showConnectionError(connerr, attempt)
 
 	if errors.Is(connerr, context.Canceled) {
+		c.Errorf("context canceled")
```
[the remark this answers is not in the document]
morestatic: the client logs aren't that noisy but for the avoidance of any doubt, i've added a touch more context.

client/client.go
morestatic: as above.

domenipavec (approved):

PR #1004, opened 2023-04-11
Add num clients and connected counts to client groups

domenipavec:
https://tracker.rport.io/issue/DEV-2825/Extend-the-client-groups-endpoint


morestatic (approved):
we should start to put some tests in place for these api fns where they are missing. now isn't the right time but, after the release, for new changes we should be considering.

PR #1012, opened 2023-04-13
DEV-2791: update attributes through API

KonradKuznicki:
DEV-2791: update attributes through API


api-doc/openapi/paths/clients_{client_id}_attributes.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - name: client_id
+      in: path
+      description: Unique client ID
+      required: true
+      schema:
+        type: string
+  requestBody:
+    description: value in the json format
+    content:
+      '*/*':
```
domenipavec, line 18: All `*/*` in this file should be `application/json`.

server/api_handler_client_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	client.SetAttributes(attributes)
+
+	err = al.clientService.GetRepo().Save(client)
+	if err != nil {
+		al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload("client attributes updated, error saving changes to local db, changes will be visible after next client connection"))
+	}
+
+	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload("ok"))
```
domenipavec, line 95: This is different from all other endpoints, where we return `StatusNoContent` if there's no actual data.

bdd/client_attributes/update/e2e_test.go
```diff
@@ -0,0 +1,99 @@
...
+	err := os.WriteFile("./client_attributes.json", []byte("{\"tags\":[\"vm\"],\"labels\":{}}"), 0600)
+	suite.NoError(err)
+	suite.ctx = context.Background()
+	ctx, cancel := context.WithTimeout(suite.ctx, time.Minute*5)
+	defer cancel()
+	suite.serverProcess, suite.clientProcess = helpers.StartClientAndServerAndWaitForConnection(ctx, suite.T(), "../../../")
+	time.Sleep(time.Millisecond * 100)
+	if suite.clientProcess.ProcessState != nil || suite.serverProcess.ProcessState != nil {
+		suite.Fail("daemons didn't start")
+	}
```
domenipavec, line 56: I think these should be part of the `StartClientAndServerAndWaitForConnection`, since it has `Wait` in a name I don't think it makes sense to have to then wait 100ms manually.

bdd/helpers/goRunner.go
```diff
@@ -7,17 +7,18 @@ import (
...
 	"os/exec"
+	"path"
 	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-func StartClientAndServerAndWaitForConnection(ctx context.Context, t *testing.T) (*exec.Cmd, *exec.Cmd) {
+func StartClientAndServerAndWaitForConnection(ctx context.Context, t *testing.T, projectRoot string) (*exec.Cmd, *exec.Cmd) {
```
domenipavec, line 17: Nit.: You could find project root by traversing up until you find `go.mod` file, then it wouldn't have to be specified manually.

client/client.go
```diff
@@ -614,6 +617,10 @@ func (c *Client) handleSSHRequests(ctx context.Context, sshClientConn *sshClient
 		var err error
 		var resp interface{}
 		switch r.Type {
+
+		case comm.RequestTypeUpdateClientAttributes:
+			c.Logger.Infof("received request to update attributes")
```
domenipavec, line 622: I don't think it makes sense to log this differently then other requests. There's a generic request type log above, if you think it should be on info level, change the level of that.

client/client.go
```diff
@@ -979,3 +986,38 @@ func (c *Client) localAddrForInterface(ifaceName string) (net.Addr, error) {
...
+		return nil, fmt.Errorf("attributes file path not set")
+	}
+
+	configHolder := &clients.Attributes{}
+	err := json.Unmarshal(payload, configHolder)
+	if err != nil {
+		return nil, fmt.Errorf("payload unreadable: %v", err)
+	}
+
+	data, err := json.Marshal(configHolder)
```
domenipavec, line 1005: Huh, why not just use payload directly?

client/client.go
```diff
@@ -979,3 +986,38 @@ func (c *Client) localAddrForInterface(ifaceName string) (net.Addr, error) {
...
+	}
+
+	c.mu.Lock()
+	c.configHolder.Client.Tags = configHolder.Tags
+	c.configHolder.Client.Labels = configHolder.Labels
+	c.mu.Unlock()
+
+	return struct {
+		Status string `json:"status"`
+	}{Status: "OK"}, nil
```
domenipavec, line 1022: I don't think you should invent response that's always the same. Just keep the response `nil`, which means ok if there's no error.

docs/content/advanced/no11-attributes.md
```diff
@@ -12,7 +12,22 @@ Clients can be described for filtering and identification by:
 - single dimension tags `["win", "server", "vm"]`
 - 2-dimensional labels  `"labels": {"country": "Germany", "city": "Cologne", "datacenter": "NetCologne GmbH" }`
 
-## Using an attributes file
+## setting attributes
+
+As of writing this Doc there are 3 ways to set up client's attributes.
+
+1. In client's `rport.conf` config file, under property tags (labels are not supported)
+2. As a separate file on the client `attributes file`
```
domenipavec, line 20: Is `attributes_file` different from `attributes_file_path` or should this be changed?

docs/content/advanced/no11-attributes.md
```diff
@@ -12,7 +12,22 @@ Clients can be described for filtering and identification by:
...
+3. Through the API - `attributes file` has to be enabled
```
domenipavec, line 21: I'm not sure what `has to be enabled` means. Is there a separate `enable` flag, or should this be `has to be set`?

client/client.go
```diff
@@ -9,11 +9,14 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"runtime"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/realvnc-labs/rport/server/clients"
```
domenipavec, line 18: I don't think we should be importing `server` package from client. By our current code organization, the shared code should live in `share`.

server/clients/client.go
```diff
@@ -539,6 +558,25 @@ func (c *Client) UserGroupHasAccessViaClientGroup(userGroups []string, allClient
...
 
+type Attributes struct {
+	Tags   []string          `json:"tags"`
+	Labels map[string]string `json:"labels"`
+}
+
+func (c *Client) GetAttributes() Attributes {
+	attr := Attributes{}
+	c.flock.RLock()
+	copierClientsToAttrs(*c, &attr) //nolint:govet
```
domenipavec, line 569: Is the copy really neccessary, wouldn't it work if you use struct embedding to add `Attributes` to `Client`?

server/clients/client.go
```diff
@@ -19,6 +21,23 @@ import (
 	"github.com/realvnc-labs/rport/share/random"
 )
 
+var copyAttrsToClient func(attributes Attributes, client *Client)
+var copierClientsToAttrs func(client Client, attributes *Attributes)
+
+func init() {
+	var err error
+
+	pairs := []dyncopy.FromToPair{dyncopy.NewPair("Tags", "Tags"), dyncopy.NewPair("Labels", "Labels")}
```
domenipavec, line 30: Is all this reflection and complicated code really necessary for 2 fields? Just make a hard coded copy function.

domenipavec (changes requested):

morestatic:
are you seeing double comments? ... if so, not sure why that's happening. i didn't submit twice.

server/api_handler_client_attributes.go
KonradKuznicki: I thought about it but, there is a situation (line 92) when there is more information sent.
But on a second thought, it would more clearly indicate the difference between line 92 and 95

docs/content/advanced/no11-attributes.md
KonradKuznicki: Yes, it is different, because we don't expect this "attributes_file" to exist even when "attributes_file_path" is set

docs/content/advanced/no11-attributes.md
KonradKuznicki: it is better explained here https://github.com/realvnc-labs/rport/pull/1012/files/b485ba0b3997fd9a2a70fabe78e584ee445adaf9#diff-825eafc99c3bb12795f2864caffa7943c13f99f7c5d17e55a3669dca4ac8ed62R91

server/clients/client.go
KonradKuznicki: @morestatic you are exactly right, I did it to start a discussion but it just so happens that after yesterday there is a ticket. Not for the full refactor (I have to confirm that with Thorsten) though at least for the client 
https://tracker.rport.io/issue/DEV-2841/Speed-up-client-search
after some hacking, I've got 50x improvement using reflection and struct coping,
from ~2.2s to ~40ms

server/clients/client.go
KonradKuznicki: @domenipavec That was my first approach, yet it would generate an enormous amount of refactoring 
which I think is necessary for multiple reasons

server/api_handler_client_attributes.go
domenipavec: @KonradKuznicki That's weird that you're returning status ok on line 92 for an error. I think it would be better to return `StatusNoContent` here and an error on line 92.

server/clients/client.go
morestatic: i think i'm with @domenipavec on this one. keep the fancy copy for the big structs and then keep this copy as simple as it can be.

client/client.go
KonradKuznicki: I agree it is ugly. I find this structure unfortunate. 
Moving just attributes to share feels wrong but moving the entire client representation feels wrong as well.
Serious refactoring sound appropriate though not in scope for this ticket.
So only code duplication seems wright but well... you somehow have to be sure you have the same structure on both ends - ugly and error-prone as well.
I think that the current solution is ugly but not error-prone.
but @morestatic @domenipavec feel free to vote

server/api_handler_client_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+func (al *APIListener) handleUpdateClientAttributes(w http.ResponseWriter, req *http.Request) {
+
+	ctx := req.Context()
+
+	client, err := al.getClientFromContext(ctx)
+	if err != nil {
+		al.jsonErrorResponseWithTitle(w, http.StatusInternalServerError, "client not present in the request")
+	}
+
+	if req.ContentLength > 2^10*5 { // limit JSON to 5KB
```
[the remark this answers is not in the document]
KonradKuznicki: to be honest it should probably be a configuration on the endpoint though - I'll research that

server/api_handler_client_attributes.go
KonradKuznicki: @domenipavec It's something between error and success because when the client will reconnect the database will be updated. I've talked with Thorsten about it and in this scenario, we are ok with "eventual consistency".

client/client.go
KonradKuznicki: I think that saving stuff on the client is quite risky (from various perspectives), this way I'm ensuring that whatever is saved should be at least readable by the client without error. 

After reading this comment and thinking deeper about the problem we should have better feature negotiation between the client and server 
because newer versions of the server might want to set up things that would be silently dismissed by the client.
Not to mention other failure modes.

Though  this would be a much bigger refactoring than this ticket

client/client.go
KonradKuznicki: I agree that nil would be enough now.
Though :
- it feels weird to return nil on the success - the rest of the code expects "something"
- it's sort of against the way Go is designed
- its harder when debugging to tell that everything is correct

But if you insist I can make it nil

client/client.go
KonradKuznicki: https://tracker.rport.io/issue/DEV-2859/Feature-version-negotiation-between-client-and-server
I've created a ticket for that

server/clients/client.go
KonradKuznicki: removed for further discussion
https://tracker.rport.io/issue/DEV-2860/refactor-manual-property-copying-between-structs

server/api_handler_client_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+
+	err = al.clientService.GetRepo().Save(client)
+	if err != nil {
+		al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload("client attributes updated, error saving changes to local db, changes will be visible after next client connection"))
+	}
+
+	al.writeJSONResponse(w, http.StatusOK, api.NewSuccessPayload("ok"))
+}
+
+func (al *APIListener) withActiveClient(next http.Handler) http.Handler {
```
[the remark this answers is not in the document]
KonradKuznicki: https://tracker.rport.io/newIssue?draftId=92-6829

client/client.go
domenipavec: I'm fine if you leave it like this, just seems a bit weird to have this hardcoded.

client/client.go
domenipavec: I think ideally all structs for server-client communication should be separate from structs we use for storing data on either client or server and stored in `share`. I'm ok if you want to leave it like this, but I think it might set a dangerous precedent for high coupling of server and client data structs.

server/clients/client.go
domenipavec: @KonradKuznicki Please also remove `dyncopy` from the codebase for now.

server/api_handler_client_attributes.go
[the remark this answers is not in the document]
domenipavec: @KonradKuznicki We already have a general `max_request_bytes` (default 10KB) for all endpoints. Do we really need a specific limit and setting for this endpoint?

server/clients/client.go
KonradKuznicki: done

server/api_handler_client_attributes.go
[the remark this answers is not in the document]
KonradKuznicki: https://tracker.rport.io/issue/DEV-2861/Replace-copy-paste-client-validation-with-client-in-context

server/api_handler_client_attributes.go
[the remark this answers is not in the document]
KonradKuznicki: I removed the limitation

client/client.go
KonradKuznicki: its' moved to shared/model/attributes

domenipavec (approved):

morestatic (approved):

PR #1017, opened 2023-04-19
[DEV-2843] Remove api session related data races

morestatic:
https://tracker.rport.io/issue/DEV-2843/Remove-api-session-related-data-races

This implementation uses struct copies rather than mutexes. What do we think about this?

### DIFF
--- a/server/api/session/cache.go
+++ b/server/api/session/cache.go
@@ -41,19 +41,18 @@ func NewCache(
 	}, nil
 }
 
-func (p *Cache) Get(ctx context.Context, sessionID int64) (*APISession, error) {
+func (p *Cache) Get(ctx context.Context, sessionID int64) (found bool, sessionInfo APISession, err error) {
 	return p.getFromCache(sessionID)
 }
 
-func (p *Cache) Save(ctx context.Context, session *APISession) (sessionID int64, err error) {
+func (p *Cache) Save(ctx context.Context, session APISession) (sessionID int64, err error) {
 	// always save to storage as we want the token last access time saved
 	sessionID, err = p.storage.Save(ctx, session)
 	if err != nil {
 		return -1, err
 	}
 
-	// make sure the session id is included in the cache version. this also updates
-	// the session ID in the supplied session.
+	// make sure the session id is included in the cache version.
 	session.SessionID = sessionID
 
 	p.cache.Set(formatID(sessionID), session, time.Until(session.ExpiresAt))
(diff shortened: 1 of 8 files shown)


server/api/session/cache.go
```diff
@@ -41,19 +41,18 @@ func NewCache(
 	}, nil
 }
 
-func (p *Cache) Get(ctx context.Context, sessionID int64) (*APISession, error) {
+func (p *Cache) Get(ctx context.Context, sessionID int64) (found bool, sessionInfo APISession, err error) {
```
domenipavec, line 44: If I understand correctly, in middleware we get the session, modify some fields and save it back. And because multiple requests can happen at the same time that was causing data races?

Operating on copies prevents the explicit data race of modifying the same object. But the underlying simultaneous read/write is still happening, except now we sort of hide it.

As long as operation that we're doing on the data doesn't depend on the previous values and we're fine with saving whatever value was later, this will work fine.
But in general I don't think this is a good approach to concurrent data access.

Also in save we're saving to storage and cache, I imagine it could happen that after one go routine saves to storage, the scheduler could switch to another that would save to both storage and cache, and then the first go routine would save to cache leading to different values in storage and cache.

morestatic: yes, the write activity is to update the last access details (time, ip address) for an api session.

i would separate the data races into logical vs golang data access. the golang data race warnings are naturally the second one. we need to be a little careful not to conflate the two. logical data races are in the application layer and may only be partially addressable via mechanisms like locking.

you're right to point out the "granularity" of the update and also the timing of the update relative to the read.

if we have separate concurrent updates of different fields then we shouldn't be copying and passing whole structs. we've probably got the update "level" wrong. whether locks with field getters/setters or smaller sub-structs are the right way to go might depend on the nature of the updates. 

if the separate concurrent updates do not dependent on updating different fields then we can using whole structs.

regarding the timing between reads and updates in concurrent scenarios, then this is in the realm of logical data races and i think it depends on the physics involved. ideally the read and update would be atomic, but quite often this isn't possible, so we have to compromise. we either live with the race because the consequences aren't serious, or we can fail the second update that's based on stale state read previously. then we must re-read and retry (which could fail again, but mostly doesn't). both approaches are valid. rport so far seems to mostly go with the former.

note with managed logical data races the data consequences are planned and understood (even if not ideal). with golang data races the outcome is unexpected, unplanned and actually the final data values are unknown.

overall, why do we care? we could just choose the locking strategy and then stick with it across all updates. my worry is that locks are pretty painful and it is easy to write code that causes deadlocks. they create a "dance" between updates where you need to make sure that you don't take a lock and then call another update fn that also takes a lock, creating a deadlock. for simple data structures this is fine but as data structures and their updates get more complex, it can get ugly fast.

there's a reason that the "functional" programming world uses immutable data as a general approach and the reason is that concurrently updating the same data is hard to reason about. better to think about any retrieved state as a snapshot and then consider staleness (of the snapshot based data) on update separately.

so ... the reason i went with the struct copies for the api sessions, was that the considerations allow it and any edge cases don't have a material impact, and therefore we don't need to consider locking the `APISession`, therefore avoiding adding more locks.

btw, i did actually implement a locking version first, and then i thought i'd try a struct copy version. so if we want to revert to a locked version then we switch can fairly quickly.

your point about the storage and cache is very valid and we'll need to decide whether requires a fix or not, depending on our assessment of the likelihood and the consequences of them being out of sync.

sorry for the essay but i wanted to have the conversation and get your input. any thoughts on the above? if we can achieve a consensus then we should apply the thinking when doing PR reviews.

domenipavec (approved):

server/api/session/cache.go
domenipavec: I think this approach is fine for this PR, but if we want to decide on a general approach we want to take with this we should discuss it with the whole team.

PR #1022, opened 2023-04-20
[DEV-2821] Add url delivery method for 2fa tokens

domenipavec:
https://tracker.rport.io/issue/DEV-2821/built-in-2FA-email-sending

### DIFF
--- a/server/api/message/url.go
+++ b/server/api/message/url.go
@@ -1,0 +1,56 @@
+package message
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+
+	email2 "github.com/realvnc-labs/rport/share/email"
+)
+
+type URLService struct {
+	senderURL string
+	baseURL   string
+}
+
+func NewURLService(senderURL, baseURL string) *URLService {
+	return &URLService{
+		senderURL: senderURL,
+		baseURL:   baseURL,
+	}
+}
+
+func (s *URLService) Send(ctx context.Context, data Data) error {
+	values := url.Values{}
+	values.Set("email", data.SendTo)
+	values.Set("token", data.Token)
+	values.Set("ttl", strconv.Itoa(int(data.TTL.Seconds())))
+	values.Set("user_agent", data.UserAgent)
+	values.Set("remote_address", data.RemoteAddress)
+	values.Set("url", s.baseURL)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, s.senderURL, strings.NewReader(values.Encode()))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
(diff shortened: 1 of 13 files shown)


server/api/message/url.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode/100 != 2 {
```
morestatic, line 44: why the `/100`? guessing there must be a reason?

server/api/message/url.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (s *URLService) Send(ctx context.Context, data Data) error {
+	values := url.Values{}
+	values.Set("email", data.SendTo)
+	values.Set("token", data.Token)
+	values.Set("ttl", strconv.Itoa(int(data.TTL.Seconds())))
+	values.Set("user_agent", data.UserAgent)
+	values.Set("remote_address", data.RemoteAddress)
+	values.Set("url", s.baseURL)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, s.senderURL, strings.NewReader(values.Encode()))
```
morestatic, line 34: this might have been also been an issue with the previous code, so not sure whether appropriate to address as part of this PR.

but i'm wondering about the user login UX. if we have a blocked connection will it appear to hang the login session? i presuming the client will timeout but if we timed out this connection before the client timeout, then the client error message might be able to be better, indicating that we were having problems sending the 2fa, rather than it just looking like rport login is being flaky?

dunno, not sure worth address yet. if we start to get feedback that an improved message would be helpful then we could tackle this then? feel free to resolve this comment unaddressed if you feel we can defer.

server/api/message/url.go
thorstenkramm: @morestatic This way, any code from 200 to 299 is fine. It's a smart way to avoid using between.

server/api/message/url.go
thorstenkramm: Valid point. I haven't seen the entire implementation, so maybe it's already solved. The 2FA sending request must use a short timeout (e.g. 2seconds) so it times out before the frontend times out.

domenipavec: @morestatic There's a `two_fa_send_timeout` in the config, that is then applied to the context before this is called. Since I'm using `NewRequestWithContext`, that timeout applies to the request being made.

server/api/message/url.go
domenipavec: This is a common way to check that the `StatusCode` is any of the `2xx` http success code.
If you have a more readable way to write this, I'd be happy to change it.

morestatic: yes, i do remember now. i have seen before. to be honest, i'd probably have a helper fn for it but certainly in nit territory, so i wouldn't worry here.

morestatic (approved):

PR #1027, opened 2023-05-04
[DEV-2842] CORS for API and http tunnels

domenipavec:
https://tracker.rport.io/issue/DEV-2842/Allow-CORS


morestatic (approved):
I've approved as the code looks fine. A couple of product-y observations that maybe @thorstenkramm can confirm whether we should consider:
1) We aren't validating the origins, which might be friendly or not? i'm not quite sure.
2) Would it be helpful to log that CORS origins are being applied? The config file is big now and it is easy to lose sight of applied config values.

PR #1101, opened 2023-06-05
dev-2928: add notification interface with mail backend

KonradKuznicki:


KonradKuznicki:
included in the alerting service

PR #1115, opened 2023-06-12
Support configurable TLS version for Caddy

jackyaz:
This PR converts the hardcode tls1.3 value in Caddy configuration to use a parameter, tls_min, to allow users to choose between tls1.2 and tls1.3 as the minimum TLS version in the same way the core rportd server/api allows.

It defaults to tls1.3 if the parameter is not present in the configuration file.


jackyaz:
Is there anything needed from me to progress this?

KonradKuznicki (approved):

PR #1164, opened 2023-07-27
Feature/dev 3048 ip address determination

thorstenkramm:
https://tracker.rport.io/issue/DEV-3048/Reliable-client-IP-address-determination


morestatic (changes requested):

morestatic:
hi @thorstenkramm noticed that there's a couple of build issues. you need to merge the latest master with your branch, so it is up to date and then the rport-plus build won't fail. then, for the failing test, it is checking the client values returned from the `GetClient` api call, and the test expectation data is missing the newly added `ext_ip_addresses` field, so that needs to be added.

client/client.go
```diff
@@ -573,7 +579,7 @@ func (c *Client) sendConnectionRequest(ctx context.Context, sshConn ssh.Conn, mi
 	c.Infof(msg)
 
 	for _, r := range remotes {
-		c.Infof("New tunnel: %s", r.String())
+		c.Infof("NewFetcher tunnel: %s", r.String())
```
morestatic, line 582: did you mean this?

morestatic (approved):
left one very small comment where looks like a search and replace (or multi-cursor edit) caught something it shouldn't.

PR #1169, opened 2023-08-03
[DEV-3026] Monitoring and Alerts Rule Testing

morestatic:
This PR introduces the alerting service rules tester API. Rather than deal with the overhead/queuing of reviewing 4 PRs, other related tickets (as below) have been merged with this PR.

https://tracker.rport.io/issue/DEV-3026/Monitoring-and-Alerts-rule-testing
https://tracker.rport.io/issue/DEV-3028/Rules-Examples
https://tracker.rport.io/issue/DEV-3043/shortcuts-for-disk-rules
https://tracker.rport.io/issue/DEV-3070/measurements-not-measures


server/client_listener.go
```diff
@@ -277,7 +293,7 @@ func (cl *ClientListener) acceptSSHConnection(w http.ResponseWriter, req *http.R
 	return sshConn, chans, reqs, clog, err
 }
 
-func (cl *ClientListener) receiveClientConnectionRequest(sshConn *ssh.ServerConn, reqs <-chan *ssh.Request, clog *logger.Logger) (connRequest *chshare.ConnectionRequest, r *ssh.Request, err error) {
+func (cl *ClientListener) receiveClientConnectionRequest(sshConn *ssh.ServerConn, reqs <-chan *ssh.Request, clog *logger.DynamicLogger) (connRequest *chshare.ConnectionRequest, r *ssh.Request, err error) {
```
KonradKuznicki, line 296: Since type is changed, maybe we could introduce interface?

server/client_listener.go
```diff
@@ -475,18 +500,28 @@ func (cl *ClientListener) handleSSHRequests(clientLog *logger.Logger, clientID s
 			continue
 
 		case comm.RequestTypePing:
-			// clientLog.Debugf("ping received from: %s", clientID)
+			var ts time.Time
+			if ClientPingsLogEnabled {
```
KonradKuznicki, line 504: could we make it more generic?

morestatic: the current logging code is mostly temporary (famous last words) just so that we're getting some timing values currently. i think how we handle long term is still open. i'd prefer just to leave this for now, and then once we're comfortable with the ssh request handling in general then we can have a more elegant solution? same for the refactoring to functions really.

server/client_listener.go
morestatic: if you're referring to the dynamic logger then it has additional methods beyond the classic logger. but again this is something i'd prefer just to leave for now as there's already a ticket for improving the dynamic logger. when we do that ticket then happy to revisit interfaces etc.

KonradKuznicki (approved):

PR #1189, opened 2023-08-14
add config options for notifications

KonradKuznicki:


server/chconfig/config.go
```diff
@@ -274,17 +276,66 @@ func (mc *MonitoringConfig) GetDataStorageDuration() (duration time.Duration) {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (n *NotificationsConfig) parseAndValidateTTL() error {
+	if len(n.LogStorageDurationString) == 0 {
+		n.LogStorageDurationString = "7d"
```
morestatic, line 304: can we move these (and below) to constants at the top of the file? or at least just above notifications config struct. having constant values hidden in fns isn't my favourite.

server/notifications/channels/scriptRunner/consumer.go
```diff
@@ -12,13 +12,15 @@ import (
 const ScriptTimeout = time.Second * 20
 
 type consumer struct {
-	l *logger.Logger
+	l   *logger.Logger
+	pwd string
```
morestatic, line 16: doesn't `pwd` stand for "print working directory"? is it the right name? just a nit really but wondering if helpful for future maintenance devs?

morestatic (approved):

morestatic (approved):

